PR #2, opened 2024-05-26
ISD-1900 initial charm with only tls relation implementation

Thanhphan1147:
Applicable spec: ISD-146(internal)

Initial implementation of the gateway-api-integrator charm. Comes with only the implementation of the certificates integration. 

The charm will block until TLS is configured, then only one action is currently supported (get-certificate) which allows debugging of the obtained certificate.

Only workflow supported at this point:
```
juju deploy ./gateway-api-integrator_ubuntu-22.04-amd64-aarch64-arm64.charm
juju deploy self-signed-certificates
juju config gateway-api-integrator service-hostname=gateway.internal
juju relate gateway-api-integrator self-signed-certificates
# At this point both application should be in active status
juju run gateway-api-integrator/0 get-certificate hostname=gateway.internal
```

### Checklist


actions.yaml
```diff
@@ -0,0 +1,8 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+get-certificate:
+  description: Returns the TLS Certificate. Intended for testing and debugging purposes.
+  params:
+    hostname:
+      type: string
+      description: Hostname to extract certs from.
```
arturo-seijas, line 8: Missing blank line here

charmcraft.yaml
```diff
@@ -1,13 +1,24 @@
...
+parts:
+  charm:
+    build-packages:
+      - build-essential
+      - python3-dev
+      - pkg-config
+      - libffi-dev
+      - libssl-dev
+      - rustc
+      - cargo
```
arturo-seijas, line 24: Same here

config.yaml
```diff
@@ -4,13 +4,12 @@
...
-      Acceptable values are: "info", "debug", "warning", "error" and "critical"
-    default: "info"
+  tls-secret-name:
+    default: ""
+    description: The name of the Kubernetes secret used to get the TLS certificate used for HTTPS. This settings overrides any tls-certificate relation defined. If left empty and no tls-certificates relation is defined, the charm will block until TLS is configured.
     type: string
+  service-hostname:
+    default: ""
+    description: The hostname of the service to create an ingress for.
+    type: string
```
arturo-seijas, line 15: Same here

metadata.yaml
```diff
@@ -1,50 +1,20 @@
...
-    # The upstream-source field is ignored by Juju. It is included here as a reference
-    # so the integration testing suite knows which image to deploy during testing. This field
-    # is also used by the 'canonical/charming-actions' Github action for automated releasing.
-    upstream-source: kennethreitz/httpbin
+  A [Juju](https://juju.is/) [charm](https://juju.is/docs/olm/charmed-operators) deploying and managing external access to HTTP/HTTPS services in a
+  Kubernetes cluster via a Gateway and a HTTPRoute resource. This requires the Kubernetes
+  cluster in question to have an Gateway API controller already deployed into it.
+summary: An operator to configure a kubernetes gateway and HTTPRoute.
+docs: https://discourse.charmhub.io/t/nginx-ingress-integrator-docs-index/4511
+issues: https://github.com/canonical/nginx-ingress-integrator-operator/issues
```
arturo-seijas, line 11: These URLs are not from this repo

requirements.txt
```diff
@@ -1 +1,4 @@
-ops >= 2.2.0
+kubernetes==24.2.0 # last version with DiscoveryV1beta1Api
+ops==2.13.0
+jsonschema==4.22.0
+cryptography==42.0.7
```
arturo-seijas, line 4: Missing blank line

src/charm.py
```diff
@@ -1,117 +1,296 @@
...
-        # Make Pebble reevaluate its plan, ensuring any services are started if enabled.
-        container.replan()
-        # Learn more about statuses in the SDK docs:
-        # https://juju.is/docs/sdk/constructs#heading--statuses
-        self.unit.status = ops.ActiveStatus()
+        tls_certificates_relation = self._tls.get_tls_relation()
+        if not tls_certificates_relation:
+            self.unit.status = WaitingStatus("Waiting for certificates relation to be created")
+            if event:
+                event.defer()
```
arturo-seijas, line 105: Have you considered not deferring here?

src/charm.py
```diff
@@ -1,117 +1,296 @@
...
+
+        Args:
+            event: The event that fires this method.
+        """
+        if not self._are_relations_ready(event):
+            return
+        hostname = self.get_hostname()
+        if not hostname:
+            self.unit.status = BlockedStatus("Waiting for hostname to be configured")
+            event.defer()
```
arturo-seijas, line 169: A config changed will be configured at some point, so you probably don't need to defer here

src/charm.py
```diff
@@ -1,117 +1,296 @@
...
+    def _on_certificate_available(self, event: CertificateAvailableEvent) -> None:
+        """Handle the TLS Certificate available event.
+
+        Args:
+            event: The event that fires this method.
+        """
+        tls_certificates_relation = self._tls.get_tls_relation()
+        if not tls_certificates_relation:
+            self.unit.status = WaitingStatus("Waiting for certificates relation to be created")
+            event.defer()
```
arturo-seijas, line 201: Analogous case here

src/charm.py
```diff
@@ -1,117 +1,296 @@
...
+    def _on_certificate_invalidated(self, event: CertificateInvalidatedEvent) -> None:
+        """Handle the TLS Certificate invalidation event.
+
+        Args:
+            event: The event that fires this method.
+        """
+        tls_certificates_relation = self._tls.get_tls_relation()
+        if not tls_certificates_relation:
+            self.unit.status = WaitingStatus("Waiting for certificates relation to be created")
+            event.defer()
```
arturo-seijas, line 268: Analogous case here

src/consts.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Constants used by the nginx-ingress-integrator charm."""
+
+CREATED_BY_LABEL = "nginx-ingress-integrator.charm.juju.is/managed-by"
```
arturo-seijas, line 6: Can't these be defined in the other files for improved cohesion?

config.yaml
Thanhphan1147: Thanks, I added the new line.

actions.yaml
Thanhphan1147: Thanks, I added the new line.

metadata.yaml
Thanhphan1147: I updated the issues and source url, I have yet create a discource page for this so I'll skip that for now

requirements.txt
Thanhphan1147: Thanks, I added the new line.

src/charm.py
```diff
@@ -1,117 +1,296 @@
...
 # Copyright 2024 Canonical Ltd.
 # See LICENSE file for licensing details.
 
-# Learn more at: https://juju.is/docs/sdk
-
-"""Charm the service.
-
-Refer to the following post for a quick-start guide that will help you
-develop a new k8s charm using the Operator Framework:
+# pylint: disable=protected-access,too-few-public-methods,too-many-lines
```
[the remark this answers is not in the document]
Thanhphan1147: removed!

src/consts.py
Thanhphan1147: Yep, I removed the consts file altogether

src/charm.py
Thanhphan1147: If we don't defer here then we'll probably need to do it in the `_reconcile` method of the charm, as this hook is only called once when TLS relation is established to create the CSR and store the resulting certificate and key

I'd prefer to keep `event.defer()` here since it makes sense to me to run this only once.

Thanhphan1147: what do you think?

arturo-seijas: Won't the handler be triggered again on `certificates_relation_created`?

Thanhphan1147: I think the workflow for both hooks is a bit different, on `certificates_relation_joined` we create the CSR and request a certificate from the certificate provider, while on `certificates_relation_created` we generate a secret and a private key and store them as juju secrets

Thanhphan1147: Not sure why it's done this way but it's the same in nginx-ingress-integrator I believe.

src/ingress_definition.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""nginx-ingress-integrator ingress definition."""
```
cbartz, line 4:
```suggestion
"""gateway-api-integrator ingress definition."""
```

tests/unit/constants.py
```diff
@@ -0,0 +1,6 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""nginx-ingress-integrator charm unit test constants."""
```
cbartz, line 4:
```suggestion
"""gateway-api-integrator charm unit test constants."""
```

src/tls_relation.py
```diff
@@ -0,0 +1,323 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+# Since the relations invoked in the methods are taken from the charm,
+# mypy guesses the relations might be None about all of them.
+"""NGINX Ingress TLS relation business logic."""
```
cbartz, line 5:
```suggestion
"""Gateway API TLS relation business logic."""
```

src/tls_relation.py
cbartz, line 1: We need to keep in mind that the duplication of code here with the nginx ingress code may require bug fixes in both charms. It may be too early for a lib for this, but just a reminder.

charmcraft.yaml
```diff
@@ -1,13 +1,24 @@
...
 # See LICENSE file for licensing details.
-# This file configures Charmcraft.
-# See https://juju.is/docs/sdk/charmcraft-config for guidance.
-
 type: charm
 bases:
   - build-on:
-    - name: ubuntu
+    - name: "ubuntu"
       channel: "22.04"
```
mthaddon, line 7: Can we target 24.04? Same for `run-on`.

config.yaml
```diff
@@ -4,13 +4,12 @@
...
-    description: |
-      Configures the log level of gunicorn. 
-
-      Acceptable values are: "info", "debug", "warning", "error" and "critical"
-    default: "info"
+  tls-secret-name:
+    default: ""
+    description: The name of the Kubernetes secret used to get the TLS certificate used for HTTPS. This settings overrides any tls-certificate relation defined. If left empty and no tls-certificates relation is defined, the charm will block until TLS is configured.
+    type: string
+  service-hostname:
```
mthaddon, line 12: Can we alphabetise the config options, so this would come before tls-secret-name. Also, I realise we're using the same approach as nginx-ingress-integrator here, but I think we should possible consider a different name here. Either `external-hostname` or `hostname`. This will probably need some discussion with a few folks before we can confirm what we want, which might be easiest outside of this PR.

config.yaml
```diff
@@ -4,13 +4,12 @@
...
-
 options:
-  # An example config option to customise the log level of the workload
-  log-level:
-    description: |
-      Configures the log level of gunicorn. 
-
-      Acceptable values are: "info", "debug", "warning", "error" and "critical"
-    default: "info"
+  tls-secret-name:
```
mthaddon, line 8: I'm not sure we need this, do we? Since we're providing the `tls-certificates` relation the charm can define this internally. I believe this option is only needed if we plan to support a different approach to managing certificates, and need to manually specify the name of the secret that will be used.

metadata.yaml
```diff
@@ -1,50 +1,20 @@
...
-  # You may remove this if your charm will run without a workload sidecar container.
-  httpbin-image:
-    type: oci-image
-    description: OCI image for httpbin
-    # The upstream-source field is ignored by Juju. It is included here as a reference
-    # so the integration testing suite knows which image to deploy during testing. This field
-    # is also used by the 'canonical/charming-actions' Github action for automated releasing.
-    upstream-source: kennethreitz/httpbin
+  A [Juju](https://juju.is/) [charm](https://juju.is/docs/olm/charmed-operators) deploying and managing external access to HTTP/HTTPS services in a
+  Kubernetes cluster via a Gateway and a HTTPRoute resource. This requires the Kubernetes
```
mthaddon, line 7: s/a HTTPRoute/an HTTPRoute/

metadata.yaml
```diff
@@ -1,50 +1,20 @@
...
+  cluster in question to have an Gateway API controller already deployed into it.
+summary: An operator to manage gateway resources.
+docs: https://discourse.charmhub.io/t/
```
mthaddon, line 10: Seems like this is a placeholder. Can we comment/remove it til we have the right URL.

src/charm.py
cbartz: I think using defer in a method which is not the event handler is not compliant with our team standard (it makes also reasoning harder):

https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619

>The only charm methods and functions that accept EventBase instances as arguments should be the event handlers themselves. In other words, every other method should not be interacting with Events. This means, among other things, that only event handlers (the methods on which Framework.observe is called) can defer events. This helps keep the execution flow simple.


 I believe the code base of nginx-ingress-integrator is older than the standard, so you may want to reconsider writing it differently.

cbartz (commented):
Looks good to me overall. Have a few minor comments. In general, I would prefer to rewrite things rather than copy from nginx-ingress.

charmcraft.yaml
Thanhphan1147: updated to 24.04, the catch is I'd have to rewrite `charmcraft.yaml` and use charmcraft 3.0.0.

config.yaml
Thanhphan1147: as discussed internally, I updated this to `external-hostname`

config.yaml
Thanhphan1147: Make sense, I don't see any real use case for `tls-secret-name` anyways so I removed that config option. Will add it again if it's asked for

metadata.yaml
Thanhphan1147: Thx, updated.

metadata.yaml
Thanhphan1147: We can't comment this out as `metadata.yaml` file is linted for this key, I'll add a comment saying that the docs page is still under construction and will be updated once ready.

src/ingress_definition.py
Thanhphan1147: updated, thx!

tests/unit/constants.py
Thanhphan1147: updated, thx!

src/tls_relation.py
Thanhphan1147: got it, thanks. I already started moving some minor fixes to the `nginx-ingress-integrator` charm, will keep that in mind as I add more implementations

src/charm.py
Thanhphan1147: Make sense, I removed the `event.defer()` from methods that are not event handlers. I still keep defer in certificate related handlers since I think it makes sense to defer there. I'm marking this and other related conversations resolved and we can move the discussion internally if needed.

src/charm.py
Thanhphan1147: I'm marking this as resolved as per my comment above.

src/charm.py
Thanhphan1147: I'm marking this as resolved as per my comment above.

src/charm.py
Thanhphan1147: I'm marking this as resolved as per my comment above.

metadata.yaml
```diff
@@ -14,6 +14,8 @@ maintainers:
   - https://launchpad.net/~canonical-is-devops
 assumes:
   - k8s-api
+  - juju >= 3.3
```
cbartz, line 17: Curious, why do we need `juju >= 3.3`?

cbartz (dismissed):
LGTM

metadata.yaml
Thanhphan1147: Since we're support noble as a run-on option, we'd need juju to be at least 3.3. On juju < 3.3 you'll get an error when specifying --base=ubuntu@24.04

.github/workflows/integration_test.yaml
```diff
@@ -5,18 +5,12 @@ on:
 
 jobs:
   integration-tests:
-    uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@main
+    uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@add_support_for_canonical_k8s
```
yanksyoon, line 8: Would there be a PR for this as well so that we can get it pointing to `main`? :D

.github/workflows/test.yaml
```diff
@@ -8,5 +8,4 @@ jobs:
     uses: canonical/operator-workflows/.github/workflows/test.yaml@main
     secrets: inherit
     with:
-      self-hosted-runner: true
-      self-hosted-runner-label: "edge"
+      self-hosted-runner: false
```
yanksyoon, line 11: May I ask if there are special requirements that prevents us from using self-hosted runner please? :D

.github/workflows/integration_test.yaml
```diff
@@ -5,18 +5,12 @@ on:
...
-      zap-target-port: 80
-      zap-rules-file-name: "zap_rules.tsv"
-      trivy-fs-enabled: true
-      trivy-image-config: "trivy.yaml"
+      provider: k8s
+      use-canonical-k8s: true
       self-hosted-runner: true
-      self-hosted-runner-label: "edge"
+      self-hosted-runner-label: "large"
+      load-test-enabled: false
```
yanksyoon, line 15: I think it is already by default set to false :D https://github.com/canonical/operator-workflows/blob/main/.github/workflows/integration_test.yaml#L60

.github/workflows/integration_test.yaml
```diff
@@ -7,9 +7,10 @@ jobs:
   integration-tests:
     uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@add_support_for_canonical_k8s
     secrets: inherit
-    runs-on: ['self-hosted', 'x64', 'noble-large']
     with:
       provider: k8s
       use-canonical-k8s: true
+      self-hosted-runner: true
+      self-hosted-runner-label: "large"
```
cbartz, line 14: Are you sure you need large?

cbartz (dismissed):

charmcraft.yaml
```diff
@@ -1,13 +1,25 @@
...
 # See LICENSE file for licensing details.
-# This file configures Charmcraft.
-# See https://juju.is/docs/sdk/charmcraft-config for guidance.
-
 type: charm
 bases:
   - build-on:
     - name: ubuntu
       channel: "22.04"
+      architectures: [amd64]
```
yanksyoon, line 8: I'm wondering if there is a need for the square brackets here :) `[]`

config.yaml
```diff
@@ -4,13 +4,8 @@
...
-  # An example config option to customise the log level of the workload
-  log-level:
-    description: |
-      Configures the log level of gunicorn. 
-
-      Acceptable values are: "info", "debug", "warning", "error" and "critical"
-    default: "info"
+  external-hostname:
+    default: ""
+    description: The hostname of the ingress.
```
yanksyoon, line 10: I might be confused here - would this be the hostname of the ingress that we want to configure it as, or the hostname of the ingress we want to supply to the charm?
```suggestion
    description: The hostname of the ingress to configure as.
```

metadata.yaml
yanksyoon: I'm wondering if we should set juju >= 3.1 since PS6 is not yet on 3.3 I believe? (I'm not so sure on this fact though) @Thanhphan1147

src/charm.py
```diff
@@ -1,117 +1,282 @@
...
+    def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
+        """Init method for the class.
 
         Args:
-            args: Arguments passed to the CharmBase parent constructor.
+            args: Variable list of positional arguments passed to the parent constructor.
         """
         super().__init__(*args)
-        self.framework.observe(self.on.httpbin_pebble_ready, self._on_httpbin_pebble_ready)
+        kubernetes.config.load_incluster_config()
```
yanksyoon, line 48: Just out of curiosity, does Juju have access to kubernetes API (I'm thinking whether it needs `--trust`). If it does need the --trust parameter, I think we might want to put this in a separate module, perhaps not during `__init__` since putting it on init will block it from registering `framework.observe`

src/charm.py
```diff
@@ -1,117 +1,282 @@
...
-        # Get a reference the container attribute on the PebbleReadyEvent
-        container = event.workload
-        # Add initial Pebble config layer using the Pebble API
-        container.add_layer("httpbin", self._pebble_layer, combine=True)
-        # Make Pebble reevaluate its plan, ensuring any services are started if enabled.
-        container.replan()
-        # Learn more about statuses in the SDK docs:
-        # https://juju.is/docs/sdk/constructs#heading--statuses
-        self.unit.status = ops.ActiveStatus()
+        return self._tls.get_tls_relation()
```
yanksyoon, line 96: I'm wondering if this returns a `bool` as a return type :)

src/charm.py
```diff
@@ -1,117 +1,282 @@
...
+
+    def _reconcile(self) -> None:
```
yanksyoon, line 98: Just out of curiosity, would this charm be big enough to have a reconcile pattern here?

src/charm.py
```diff
@@ -1,117 +1,282 @@
...
+                {
+                    f"certificate-{hostname}": tls_rel_data[f"certificate-{hostname}"],
+                    f"ca-{hostname}": tls_rel_data[f"ca-{hostname}"],
+                    f"chain-{hostname}": tls_rel_data[f"chain-{hostname}"],
                 }
-            },
-        }
+            )
+        else:
+            event.fail("Certificate not available")
```
yanksyoon, line 136: Just a nitpick:
If we do 
```
if not <condition>:
    event.fail()
    return
```
first, we get to reduce nesting for aesthetics :D

src/charm.py
```diff
@@ -1,117 +1,282 @@
...
+
+    def _on_certificates_relation_created(self, event: RelationCreatedEvent) -> None:
+        """Handle the TLS Certificate relation created event.
+
+        Args:
+            event: The event that fires this method.
+        """
+        if not self._are_relations_ready():
+            self.unit.status = WaitingStatus("Waiting for certificates relation to be created")
+            event.defer()
```
yanksyoon, line 146: I think this event does not need to be deferred, since if the relations become ready it will get another separate event.

.github/workflows/integration_test.yaml
Thanhphan1147: I can't bootstrap canonical k8s with edge runners due to them not having enough disk space (maybe canonical k8s controllers have a higher resource requirements than microk8s ones). So I skipped medium since I'd add more tests in the future.

src/charm.py
yanksyoon: Sorry if I've missed this comment - please disregard my above comments on defer if the same applies. (The Github changed files mode didn't quite let me see the prev comments)

src/charm.py
```diff
@@ -1,117 +1,282 @@
...
+                try:
+                    secret = self.model.get_secret(label=f"private-key-{hostname}")
+                    secret.remove_all_revisions()
+                except SecretNotFoundError:
+                    LOGGER.warning("Juju secret for %s already does not exist", hostname)
+                    continue
+            try:
+                self._tls.pop_relation_data_fields(
+                    [f"key-{hostname}", f"password-{hostname}"],
+                    tls_certificates_relation,  # type: ignore[arg-type]
```
yanksyoon, line 232: Nitpick: would type cast be helpful here?

src/charm.py
```diff
@@ -1,117 +1,282 @@
...
+        tls_relation = self._tls.get_tls_relation()
+        if tls_relation:
+            tls_relation.data[self.app].clear()
+        if JujuVersion.from_environ().has_secrets:
+            hostname = self.get_hostname()
+            try:
+                secret = self.model.get_secret(label=f"private-key-{hostname}")
+                secret.remove_all_revisions()
+            except SecretNotFoundError:
+                LOGGER.warning("Juju secret for %s already does not exist", hostname)
```
yanksyoon, line 278: Just out of curiosity, may I ask when the secret may already be gone?

src/gateway_definition.py
yanksyoon, line 1: I'm wondering if the contents of this file might be better suitable as a charm state, since this module is coupled with the charm :)

yanksyoon (commented):
Reviewed up to gateway_definition.py, thank you!

.github/workflows/integration_test.yaml
Thanhphan1147: Yep, we still need to have a discussion internally on how to properly implement this on main, since there's also an option of adding support for canonical k8s directly in `charmed-kuberentes/action-operator` in which case this patch won't be needed anymore

Thanhphan1147: So I'll leave it as it is for now :D

.github/workflows/test.yaml
Thanhphan1147: Not a special requirement but I did this to reduce the load on our self-hosted runners as this workflow is quite lightweight :smile:

charmcraft.yaml
Thanhphan1147: I'll still need to decide whether we should support multiple archs here, and also there's this point of moving to charmcraft v3 which will change this completely, so I think we can leave it like this for now

src/charm.py
Thanhphan1147: good point, most of the tls logic was copied over from the nginx-ingress-integrator charm, so things like this can be overlooked!

I'll refactor this

src/gateway_definition.py
Thanhphan1147: I'm planning to use it as a state, since we'll need a dataclass to store the definition of the gateway resource. But I wouldn't necessarily rename this file to `state.py` as I'd like to have modular charm state that spread across a few files like what we've discussed internally.

src/charm.py
Thanhphan1147: The scope would be close to nginx-ingress-integrator so I'd say i can get quite complex (the scope can grow even bigger too since gateway API offers more features than nginx ingress)

src/charm.py
Thanhphan1147: Might be when it's manually deleted by the operator via `juju remove-secret`

src/charm.py
Thanhphan1147: I'm not sure I see where could a type cast can be used, this method simply removes the 2 keys from the relation databag. Can you elaborate a bit on this?

.github/workflows/integration_test.yaml
Thanhphan1147: Right, I'm removing it. Thanks!

config.yaml
Thanhphan1147: This will be the hostname we want to supply to the ingress. As a first step we'll implement a path routing similar to the `traefik-k8s` charm where a hostname is configured on the gateway and the url of the application will be something like `https://<hostname>/<path>`

src/charm.py
Thanhphan1147: only for making API calls I think, here it's just loading the service account assigned to the charm pod so no error handling is needed here I think.

src/charm.py
Thanhphan1147: refactored!

src/charm.py
Thanhphan1147: indeed, though this code still seems to work :smile:. I changed it to make sure that a bool is returned

src/charm.py
Thanhphan1147: This is similar to @arturo-seijas's comment about deferral, though in this case `_on_relation_created` and `_on_relation_joined` are doing 2 different things, while the former generates a private key and password and creates a juju secret, the latter creates the CSR from the juju secret and sends it back to the TLS provider.

Thanhphan1147: I think we can refactor these 2 event handlers so that `defer` is no longer needed, but I would prefer to do it in a follow-up PR.

src/charm.py
```diff
@@ -124,17 +124,17 @@ def _on_get_certificate_action(self, event: ActionEvent) -> None:
...
-            event.set_results(
-                {
-                    f"certificate-{hostname}": tls_rel_data[f"certificate-{hostname}"],
-                    f"ca-{hostname}": tls_rel_data[f"ca-{hostname}"],
-                    f"chain-{hostname}": tls_rel_data[f"chain-{hostname}"],
-                }
-            )
-        else:
+        if not tls_rel_data.get(f"certificate-{hostname}"):
             event.fail("Certificate not available")
```
cbartz, line 128:
```suggestion
        if not tls_rel_data.get(f"certificate-{hostname}"):
            event.fail("Certificate not available")
            return
```

Thanhphan1147: Added the fix, thanks!

cbartz (dismissed):

src/charm.py
```diff
@@ -1,117 +1,278 @@
...
+        tls_rel_data = tls_certificates_relation.data[self.app]
+        if not tls_rel_data.get(f"certificate-{hostname}"):
+            event.fail("Certificate not available")
+            return
 
-        Change this example to suit your needs. You'll need to specify the right entrypoint and
-        environment configuration for your specific workload.
+        event.set_results(
+            {
+                f"certificate-{hostname}": tls_rel_data[f"certificate-{hostname}"],
```
amandahla, line 133: Minor: check if the relation data exists before setting the results?

src/charm.py
```diff
@@ -1,117 +1,278 @@
...
-        # https://juju.is/docs/sdk/constructs#heading--statuses
-        self.unit.status = ops.ActiveStatus()
+        if not self._are_relations_ready():
+            self.unit.status = WaitingStatus("Waiting for certificates relation to be created")
+            event.defer()
+            return
+        hostname = self.get_hostname()
+        if not hostname:
+            self.unit.status = BlockedStatus("Waiting for hostname to be configured")
+            event.defer()
```
amandahla, line 152: I saw Arturo's comment and I believe in this case the defer is not needed, right? Since a config changed will be received after setting hostname.

src/charm.py
```diff
@@ -1,117 +1,278 @@
...
-        this method.
+        Args:
+            event: The event that fires this method.
+        """
+        if not self._are_relations_ready():
+            self.unit.status = WaitingStatus("Waiting for certificates relation to be created")
+            event.defer()
+            return
+        hostname = self.get_hostname()
+        if not hostname:
```
amandahla, line 167: I believe this could be for a future PR but using charm_state + validator could make it easier to guarantee that hostname is a required configuration. Synapse has the same case for server_name.

src/charm.py
Thanhphan1147: That's a good point, thanks! I updated the check to check for all relevant keys in the relation data. I also updated to using `.get()` to avoid KeyError. Will update this on the nginx ingress integrator side as well

src/charm.py
```diff
@@ -1,117 +1,278 @@
...
-        container = event.workload
-        # Add initial Pebble config layer using the Pebble API
-        container.add_layer("httpbin", self._pebble_layer, combine=True)
-        # Make Pebble reevaluate its plan, ensuring any services are started if enabled.
-        container.replan()
-        # Learn more about statuses in the SDK docs:
-        # https://juju.is/docs/sdk/constructs#heading--statuses
-        self.unit.status = ops.ActiveStatus()
+        if not self._are_relations_ready():
+            self.unit.status = WaitingStatus("Waiting for certificates relation to be created")
```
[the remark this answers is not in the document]
Thanhphan1147: Since this is running in the `_on_certificates_relation_created` hook I think we can be sure that the relation will be ready at some point, so nothing for the user to do here. We do block the charm here: https://github.com/canonical/gateway-api-integrator-operator/blob/a40b631991d954e5afdf5a349766e0677ec1f716/src/charm.py#L101-L102 since at this point it's the operator's action to create a relation.

src/charm.py
Thanhphan1147: I keep `defer` here because I think this handler should only run once. Here it's creating the juju secret that will be used to create the CSR. If we remove `defer` here then we'd need to move also the event handler code into the main `reconcile` loop. which might make it a bit too complex with the k8s resource management logic that we'd also want to put there.

What do you think? should we keep it like this or should we remove `defer` and move the below code into the main `reconcile` loop?

src/charm.py
Thanhphan1147: good point, I'll add it to a followup PR.

src/charm.py
```diff
@@ -124,15 +124,18 @@ def _on_get_certificate_action(self, event: ActionEvent) -> None:
...
             return
 
         event.set_results(
             {
-                f"certificate-{hostname}": tls_rel_data[f"certificate-{hostname}"],
-                f"ca-{hostname}": tls_rel_data[f"ca-{hostname}"],
-                f"chain-{hostname}": tls_rel_data[f"chain-{hostname}"],
+                f"certificate-{hostname}": tls_rel_data.get(f"certificate-{hostname}"),
+                f"ca-{hostname}": tls_rel_data.get(f"ca-{hostname}"),
+                f"chain-{hostname}": tls_rel_data.get(f"chain-{hostname}"),
```
cbartz, line 138: I think the above check for data availability makes the use of `.get` here unnecessary.

cbartz (approved):

src/charm.py
Thanhphan1147: Yes, but I think we should still use `.get()` instead of `[]` since it's objectively safer since it's unconditional and never raises an Exception

cbartz: Gotcha, but what is the downside of an exception here? If we expect the keys to exist (which you do with the above check), I would rather have an exception thrown if a key is missing than have invalid relation data which is much harder to debug.

cbartz: Gotcha, but what is the downside of an exception here? If we expect the keys to exist (which you do with the above check), I would rather have an exception thrown if a key is missing than return invalid data, as this alerts us to invalid programming logic.

src/charm.py
amandahla: Hm...I see your point.
Well, if in the future you can guarantee that the hostname will always be set (that charm state + validator comment somewhere), then this could be removed right?

Thanhphan1147: yes, I already have a PR (a bit down the line since it's like 4 PRs pending before it) to refactor the charm state logic and add pydantic validation so that it can be removed.

src/charm.py
Thanhphan1147: I wouldn't say a down side, but rather added complexity in that we'd need to add exception handling to  the potential `KeyError` that might be thrown, which will seem like added code for not much value. And we'd probably have to add unit test for coverage as well.

amandahla (approved):
LGTM

src/charm.py
Thanhphan1147: I discussed about this internally and I'll revert back to using `[]`, the idea is indeed to let the exception flow and alert us of that really strange situations. I'll do it in a follow-up PR.

src/charm.py
Thanhphan1147: I'm closing this conversation for now, we can continue our discussion internally if needed

PR #3, opened 2024-05-31
ISD-1973 Implement gateway resource manager with HTTP listener

Thanhphan1147:
Applicable spec: ISD-146(internal)

### Overview
This PR adds the following:
* use lightkube as the kubernetes client
* A gateway resource definition class: this is a component of charmstate that defines the k8s resource to be created.
* A generic resource manager class to manage kubernetes resources.
* A gateway resource manager class to manage gateway resource
* A new config option gateway-class (Required) that allows customizing the kubernetes gateway class. Currently no validation is done on this config except for non-null validation.

Use case:
On canonical-k8s, run:
```
# in the gateway model
juju deploy self-signed-certificates
juju deploy gateway-api-integrator --config external-hostname=hostname --config gateway-class=cilium
juju relate gateway-api-integrator self-signed-certificates

kubectl get gateway -n gateway
```
Should show the gateway being created

Notes:
1. Patching is done using Server-Side Apply, during init the charm sets `field_manager=self.app.name` and when patching it always use `force=True`, this ensures that the charm stays in control of the resource.
2. Gateway address is reported with juju status at the end of the reconciliation loop.
### Checklist


.github/workflows/test.yaml
```diff
@@ -5,7 +5,8 @@ on:
 
 jobs:
   unit-tests:
-    uses: canonical/operator-workflows/.github/workflows/test.yaml@main
+    uses: canonical/operator-workflows/.github/workflows/test.yaml@remove_default_jammy_label
```
amandahla, line 8: Just to be sure: is expected to use/merge using this branch, right?

src/charm.py
```diff
@@ -96,12 +95,54 @@ def _are_relations_ready(self) -> bool:
...
+            labels=self._labels,
+            client=client,
+        )
+
+        try:
+            gateway = gateway_resource_manager.define_resource(gateway_resource_definition)
+        except (CreateGatewayError, InvalidResourceError) as exc:
+            LOGGER.error("Error creating the gateway resource %s", exc)
+            raise RuntimeError("Cannot create gateway.") from exc
+        except KuberentesCreateResourceError as exc:
```
amandahla, line 136: s/Kuberentes/Kubernetes

What the user should do in this case once the charm is blocked?

src/charm.py
```diff
@@ -96,12 +95,54 @@ def _are_relations_ready(self) -> bool:
...
+            self.unit.status = BlockedStatus(exc.msg)
+            return
+
+        self.unit.status = WaitingStatus("Waiting for gateway address")
+        if gateway_address := gateway_resource_manager.gateway_address(gateway.metadata.name):
+            self.unit.status = ActiveStatus(f"Gateway addresses: {gateway_address}")
+        else:
+            self.unit.status = WaitingStatus("Gateway address unavailable")
```
amandahla, line 144: Is there any action that should be taken here?

src/charm.py
```diff
@@ -166,12 +211,15 @@ def _on_certificates_relation_joined(self, event: RelationJoinedEvent) -> None:
...
             return
-        hostname = self.get_hostname()
-        if not hostname:
-            self.unit.status = BlockedStatus("Waiting for hostname to be configured")
-            event.defer()
+        try:
+            gateway_resource_definition = GatewayResourceDefinition.from_charm(self)
+        except InvalidCharmConfigError as exc:
+            LOGGER.error("Invalid charm config: %s", exc.msg)
+            self.unit.status = BlockedStatus("Invalid charm configuration")
```
amandahla, line 218: Minor: maybe adding each configuration is invalid in the message or pointing to check the logs.

.github/workflows/test.yaml
Thanhphan1147: Also because I'm not sure how to best proceed in this situation, maybe we can do a more complex logic and check if the labels has a release different than jammy, use that. Otherwise set jammy by default.

It's still unclear to me so I think keeping it like this for this PR is fine for now

src/charm.py
Thanhphan1147: I believe exc.msg will tell the user to do juju trust --scope=cluster.

src/resource_definition.py
```diff
@@ -0,0 +1,85 @@
...
+    """Charm configuration.
+
+    Attrs:
+        gateway_class (_type_): _description_
+        external_hostname: The configured gateway hostname.
+    """
+
+    gateway_class: str = Field(min_length=1)
+    external_hostname: str = Field(
+        pattern=r"[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*"
```
amandahla, line 40: This is the same as the one used by NGINX charm, right? I wonder if wouldn't be nice to have some kind of library with all the regex that we are using

src/charm.py
Thanhphan1147: If the gateway address can't be fetched then that means k8s didn't assign the gateway and IP address / endpoint. Unfortunately since the charm doesn't know how k8s is setup, or even what the gateway controller is, it can't offer much help here.

src/charm.py
Thanhphan1147: It should already be in exc.msg but I overlooked this one, will update it. Thanks!

src/resource_manager/gateway.py
```diff
@@ -0,0 +1,238 @@
...
+        Args:
+            name (str): name of the gateway resource.
+
+        Returns:
+            Optional[str]: The addresses assigned to the gateway object, or none.
+        """
+        deadline = time.time() + 60
+        delay = 5
+        gateway_address = None
+        while time.time() < deadline:
```
amandahla, line 222: Shouldnt we use some kind of library to do that retry flow?
@gregory-schiano 
Postgresql is using tenacity
https://github.com/canonical/postgresql-k8s-operator/blob/04c217e70b2e41d5a68fdc604404edc3dcb0d5b7/src/upgrade.py#L107

src/resource_definition.py
Thanhphan1147: Yep, a lib implementation is in the backlog i think, but it'll be way down the line

src/resource_manager/resource_manager.py
```diff
@@ -0,0 +1,252 @@
...
+    ) -> None:
+        """Remove unused resources.
+
+        Args:
+            exclude: The name of resource to be excluded from the cleanup.
+        """
+        for resource in self._list_resource():
+            excluded_resource_name = resource_name(exclude)
+            res_name = resource_name(resource)
+            if res_name is None or excluded_resource_name is None:
```
yanksyoon, line 242: could `res_name` and `excluded_resource_name` be empty strings, i.e. "" ?
This could be `if not res_name or not excluded_resource_name` :)
I'm asking because I'm wondering if `None` represents a special case

tests/integration/test_certificates_integration.py
```diff
@@ -12,6 +12,7 @@
 
 logger = logging.getLogger(__name__)
 TEST_EXTERNAL_HOSTNAME_CONFIG = "gateway.internal"
+GATEWAY_CLASS_CONFIG = "cilium"
```
yanksyoon, line 15: It seems that GATEWAY_CLASS_CONFIG and TEST_EXTERNAL_HOSTNAME_CONFIG  is also declared on tests/integration/conftest.py, would it be possible to share this constant?

tests/unit/test_certificates_integration.py
```diff
@@ -50,7 +52,7 @@ def test_cert_relation(
     get_secret_mock = MagicMock()
     monkeypatch.setattr(ops.model.Model, "get_secret", get_secret_mock)
 
-    harness.update_config({"external-hostname": "igress-internal"})
+    harness.update_config({"external-hostname": "igress-internal", "gateway-class": "cilium"})
```
yanksyoon, line 55: Could we use the defined constant here?

yanksyoon (commented):
Very nice! Minor comments & nitpicks, thank you!

src/resource_manager/resource_manager.py
```diff
@@ -0,0 +1,252 @@
...
+        try:
+            return func(*args, **kwargs)
+        except ApiError as exc:
+            if exc.status.code == 403:
+                logger.error(
+                    "Insufficient permissions to create the k8s service, "
+                    "will request `juju trust` to be run"
+                )
+                juju_trust_cmd = "juju trust <gateway-api-integrator> --scope=cluster"
+                raise KuberentesCreateResourceError(
```
amandahla, line 103: If the exception is raised only for 403 statuscode, couldn't be something more related like UnauthorizedAccessException or similar?

src/resource_manager/resource_manager.py
```diff
@@ -0,0 +1,252 @@
...
+
+        Returns:
+            The name of the created or modified resource.
+
+        Raises:
+            InvalidResourceError: If the generated resource is invalid.
+        """
+        resource_list = self._list_resource()
+        resource = self._gen_resource_from_definition(definition)
+        if not resource.metadata:
```
amandahla, line 205: Couldnt the helper function resource_name be used here?

src/charm.py
amandahla: Hm...but if its waitingstatus, could be changed/solved by some other event? Or if not, should be an error?

.github/workflows/test.yaml
cbartz: @Thanhphan1147  You might want to create a PR on operator workflows and we can discuss this there. It probably makes sense to remove the hardcoded jammy label and replace it with something like `inputs.self-hosted-runner-image` with a default of jammy.

tests/unit/test_certificates_integration.py
mthaddon: Also, I think `igress-internal` should be `ingress-internal`.

src/resource_manager/resource_manager.py
Thanhphan1147: yes indeed, I changed this to using the `resource_name` helper. Thanks!

tests/unit/test_certificates_integration.py
Thanhphan1147: moved the constant over to this file.

tests/integration/test_certificates_integration.py
Thanhphan1147: yep, I imported the constant and used them instead.

src/resource_manager/gateway.py
```diff
@@ -0,0 +1,238 @@
...
+    create_namespaced_resource,
+)
+from lightkube.models.meta_v1 import ObjectMeta
+from lightkube.types import PatchType
+
+from resource_definition import GatewayResourceDefinition
+
+from .resource_manager import ResourceManager, _map_k8s_auth_exception
+
+LOGGER = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
Thanhphan1147: good point, I removed all upper case variables.

src/resource_definition.py
```diff
@@ -0,0 +1,85 @@
...
+                gateway_class=typing.cast(str, charm.config.get("gateway-class")),
+                external_hostname=typing.cast(str, charm.config.get("external-hostname")),
+            )
+            namespace = charm.model.name
+            gateway_name = charm.app.name
+        except ValidationError as exc:
+            error_fields = set(
+                itertools.chain.from_iterable(error["loc"] for error in exc.errors())
+            )
+            error_field_str = " ".join(f"{f}" for f in error_fields)
```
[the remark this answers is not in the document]
Thanhphan1147: thanks, fixed.

src/charm.py
```diff
@@ -77,15 +81,10 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
-        """Get a list containing all ingress hostnames.
-
-        Returns:
-            A list containing service and additional hostnames
-        """
-        # The relation will always exist when this method is called
-        external_hostname = cast(str, get_config(self, "external-hostname"))
-        return external_hostname
+    @property
+    def _labels(self) -> typing.Dict[str, str]:
```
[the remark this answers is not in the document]
Thanhphan1147: thanks for the tip, I refactored imports and only use `typing.`

requirements.txt
```diff
@@ -1,4 +1,5 @@
-kubernetes==30.1.0 # last version with DiscoveryV1beta1Api
+lightkube==0.15.2
 ops==2.14.0
 jsonschema==4.22.0
 cryptography==42.0.8
+pydantic>=1,<2
```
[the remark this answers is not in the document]
Thanhphan1147: added the exact version pin.

src/charm.py
```diff
@@ -96,12 +95,54 @@ def _are_relations_ready(self) -> bool:
...
+        except ConfigError as exc:
+            LOGGER.error("Error initializing the lightkube client: %s", exc)
+            raise RuntimeError("Error initializing the lightkube client.") from exc
+
+        try:
+            gateway_resource_definition = GatewayResourceDefinition.from_charm(self)
+        except InvalidCharmConfigError as exc:
+            LOGGER.error("Invalid charm config: %s", exc.msg)
+            self.unit.status = BlockedStatus("Invalid charm configuration")
+            return
```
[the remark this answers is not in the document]
Thanhphan1147: Good point, I have a follow-up PR pending to address charm state and configuration checks : https://github.com/canonical/gateway-api-integrator-operator/pull/5 so I'll make the changes there to be more in line with the purpose of the PR.

I'll mark this conversation as closed for now

src/resource_manager/gateway.py
Thanhphan1147: I created a ticket for this (ISD-2002 - internal). Will make a follow up PR for it as I agree that a library would work best for this. I'm closing this conversation and we can move our discussion internally if needed.

src/charm.py
Thanhphan1147: Either way works imho, the charm doesn't necessarily need to go into Error state as it's still functional (the gateway resource was created successfully in this case and tls service is working fine too). On the other hand if the k8s cluster is misconfigured in someway then an operator can reconfigure the cluster so that a LB ip address is assigned to the charm, in which case we'd want a blocked status. But the k8s cluster can also be taking some time to assign the IP address, in which case there's nothing to be done aside from wait for the cluster to do its job. 

All in all I think a waiting status is a good middle ground here, what do you think?

src/charm.py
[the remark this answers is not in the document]
Thanhphan1147: will need some further refactoring and internal discussion if we want to follow this approach for all of the charm's module, so I'm marking this as resolved for now.

src/resource_manager/resource_manager.py
Thanhphan1147: good point, I was keeping it generic in case we'd want to use it for other 4xx return codes, but I changed the name to `InsufficientPermissionError`

src/resource_manager/resource_manager.py
Thanhphan1147: good point, empty `res_name` should never be the case but it's indeed a possibility we need to cover, I changed all relevant sections to `not res_name`

src/resource_manager/resource_manager.py
```diff
@@ -0,0 +1,252 @@
...
+            raise InvalidResourceError("Missing resource metadata.")
+
+        if not resource.metadata.name:
+            raise InvalidResourceError("Missing resource name.")
+
+        resources = [r.metadata.name for r in resource_list if r.metadata]
```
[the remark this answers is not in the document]
Thanhphan1147: I updated this to avoid the confusion.

src/resource_manager/resource_manager.py
```diff
@@ -0,0 +1,252 @@
...
+def resource_name(resource: AnyResource | None) -> typing.Optional[str]:
+    """Get the resource name from metadata.
+
+    Args:
+        resource (AnyResource): The kubernetes resource.
+
+    Returns:
+        typing.Optional[str]: The resource name, or None if not set.
+    """
+    if resource is None:
```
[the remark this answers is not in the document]
Thanhphan1147: Can you elaborate a bit on this? Looking at the suggestion I think that 
`return resource and resource.metadata and resource.metadata.name` would give us a bool but we'd want to return the resource name (str) here

src/resource_manager/gateway.py
```diff
@@ -0,0 +1,238 @@
...
+
+        if not gateway_classes:
+            LOGGER.error("Cluster has no available gateway class.")
+            raise CreateGatewayError("No gateway class available.")
+
+        gateway_class_names = [
+            gateway_class.metadata.name
+            for gateway_class in gateway_classes
+            if gateway_class.metadata
+        ]
```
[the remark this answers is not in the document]
Thanhphan1147: good point, I changed to tuple to be a bit better.

src/charm.py
```diff
@@ -96,12 +95,54 @@ def _are_relations_ready(self) -> bool:
...
+        gateway_resource_manager = GatewayResourceManager(
+            namespace=gateway_resource_definition.namespace,
+            labels=self._labels,
+            client=client,
+        )
+
+        try:
+            gateway = gateway_resource_manager.define_resource(gateway_resource_definition)
+        except (CreateGatewayError, InvalidResourceError) as exc:
+            LOGGER.error("Error creating the gateway resource %s", exc)
```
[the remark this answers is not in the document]
Thanhphan1147: I don't really see `logger.exception` being used that much in our charms, so just out of curiosity, would we always use `logger.exception` or only in specific cases?

src/charm.py
```diff
@@ -96,18 +94,60 @@ def _are_relations_ready(self) -> bool:
...
+            or when creating the gateway resource fails.
+        """
+        try:
+            # Set field_manager for server-side apply when patching resources
+            # Keep this consistent across client initializations
+            kubeconfig = KubeConfig.from_service_account()
+            client = Client(config=kubeconfig, field_manager=self.app.name)
+        except ConfigError as exc:
+            logger.error("Error initializing the lightkube client: %s", exc)
+            raise RuntimeError("Error initializing the lightkube client.") from exc
```
yanksyoon, line 110: I think `logger.exception` might be enough here to print the exception stack trace :D
https://docs.python.org/3/library/logging.html#logging.Logger.exception

yanksyoon: And other `logger.error`s from exceptions :)

src/resource_manager/gateway.py
```diff
@@ -0,0 +1,238 @@
...
+        )
+
+    @_map_k8s_auth_exception
+    def _list_resource(self) -> List[GenericNamespacedResource]:
+        """List gateway resources in a given namespace based on a label selector.
+
+        Returns:
+            A list of matched gateway resources.
+        """
+        return list(
```
yanksyoon, line 191: Out of curiosity, I'm wondering if wrapping the value in list from `self._client.list` is needed :D (i'm guessint list() would return a list)

tests/unit/test_certificates_integration.py
```diff
@@ -12,24 +12,28 @@
...
 
+from .conftest import GATEWAY_CLASS_CONFIG, TEST_EXTERNAL_HOSTNAME_CONFIG
+
 
-@pytest.mark.usefixtures("patch_load_incluster_config")
 def test_generate_password(harness: Harness):
     """
     arrange: given a charm with no connectable container.
     act: when agent relation joined event is fired.
     assert: the event is deferred.
```
yanksyoon, line 22: I'm wondering if the docstrings for this tests should be updated :)

tests/unit/test_certificates_integration.py
```diff
@@ -12,24 +12,28 @@
...
 
-@pytest.mark.usefixtures("patch_load_incluster_config")
 def test_cert_relation(
-    harness: Harness, certificates_relation_data: Dict[str, str], monkeypatch: pytest.MonkeyPatch
+    harness: Harness,
+    certificates_relation_data: Dict[str, str],
+    monkeypatch: pytest.MonkeyPatch,
 ):
     """
     arrange: given a charm with no connectable container.
```
yanksyoon, line 39: for this docstring too :D

yanksyoon (dismissed):
LGTM, minor nitpicks :)

src/charm.py
Thanhphan1147: Good idea, I'll add it in a follow-up PR.

src/resource_manager/gateway.py
Thanhphan1147: Yes, I had to add `list()` for the linter to pass since apparently `client.list()` returns an iterator object rather than a list (which is indeed not intuitive at all :sweat_smile: )

tests/unit/test_certificates_integration.py
Thanhphan1147: yep, I updated the docstring!

tests/unit/test_certificates_integration.py
Thanhphan1147: yep, I updated the docstring for this one too.

src/resource_manager/resource_manager.py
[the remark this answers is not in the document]
Thanhphan1147: I see, that's nice and clean, I updated the code here to `return resource and resource.metadata and resource.metadata.name`

src/charm.py
Thanhphan1147: got it, I updated from `logger.error` to `logger.exception` in relevant places.

src/charm.py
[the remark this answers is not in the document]
Thanhphan1147: I updated from logger.error to logger.exception in relevant places.

src/resource_manager/resource_manager.py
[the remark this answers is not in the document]
Thanhphan1147: mypy is not happy with this though, I'll revert and discuss it further internally, will create a follow-up PR if needed.

```
int: commands[5]> mypy /home/tphan025/Canonical/gateway-api-integrator-operator/src/ /home/tphan025/Canonical/gateway-api-integrator-operator/tests/
src/resource_manager/resource_manager.py:121: error: Incompatible return value type (got "Endpoints | ObjectMeta | str | None", expected "str | None")  [return-value]
src/resource_manager/resource_manager.py:121: error: Incompatible return value type (got "EndpointSlice | ObjectMeta | str | None", expected "str | None")  [return-value]
src/resource_manager/resource_manager.py:121: error: Incompatible return value type (got "Service | ObjectMeta | str | None", expected "str | None")  [return-value]
src/resource_manager/resource_manager.py:121: error: Incompatible return value type (got "Secret | ObjectMeta | str | None", expected "str | None")  [return-value]
src/resource_manager/resource_manager.py:121: error: Incompatible return value type (got "GenericNamespacedResource | ObjectMeta | str | None", expected "str | None")  [return-value]
```

.github/workflows/test.yaml
Thanhphan1147: Got it, I will create a PR for operator workflows and let's discuss there.

Thanhphan1147:
Hi @gregory-schiano, when you have the time can you kindly review this PR? Thanks!

yanksyoon (approved):
LGTM, thanks for the quick changes :)

amandahla (approved):
LGTM

.github/workflows/test.yaml
Thanhphan1147: PR opened : https://github.com/canonical/operator-workflows/pull/345

requirements.txt
```diff
@@ -1,4 +1,5 @@
-kubernetes==30.1.0 # last version with DiscoveryV1beta1Api
+lightkube==0.15.2
 ops==2.14.0
 jsonschema==4.22.0
 cryptography==42.0.8
+pydantic==1.10.15
```
gregory-schiano, line 5: Why using pydantic 1 ? We should go for pydantic 2

src/charm.py
```diff
@@ -96,18 +94,60 @@ def _are_relations_ready(self) -> bool:
...
 
     def _reconcile(self) -> None:
-        """Reconcile charm status based on configuration and integrations."""
+        """Reconcile charm status based on configuration and integrations.
+
+        Raises:
+            RuntimeError: when initializing the lightkube client fails,
+            or when creating the gateway resource fails.
+        """
+        try:
```
gregory-schiano, line 103: Are you able to check if the charm was deployed using `--trust` via Juju or Ops ?

src/charm.py
```diff
@@ -96,18 +94,60 @@ def _are_relations_ready(self) -> bool:
...
+            # Set field_manager for server-side apply when patching resources
+            # Keep this consistent across client initializations
+            kubeconfig = KubeConfig.from_service_account()
+            client = Client(config=kubeconfig, field_manager=self.app.name)
```
gregory-schiano, line 107: You can set the namespace already there and it will be used for any resource action, so you won't need it anymore in your resource controllers

src/resource_manager/resource_manager.py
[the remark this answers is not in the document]
gregory-schiano: Perhaps:
```python
if resource is None or resource.metadata is None or resource.metadata.name is None:
  return None
return resource.metadata.name
```
or
```python
if resource and resource.metadata and resource.metadata.name:
  return resource.metadata.name
return None
```
(return None could even be avoided as the method returns None by default)

src/resource_manager/resource_manager.py
```diff
@@ -0,0 +1,250 @@
...
+        """
+
+    @property
+    @abc.abstractmethod
+    def _namespace(self) -> str:
+        """Abstract property that returns the namespace of the controller.
+
+        Returns:
+            The namespace.
+        """
```
gregory-schiano, line 149: If the namespace is passed via the client, this become unecessary

src/resource_manager/resource_manager.py
```diff
@@ -0,0 +1,250 @@
...
+        Raises:
+            InvalidResourceError: If the generated resource is invalid.
+        """
+        resource_list = self._list_resource()
+        resource = self._gen_resource_from_definition(definition)
+        res_name = resource_name(resource)
+        if not res_name:
+            raise InvalidResourceError("Missing resource name.")
+
+        resources = [resource_name(r) for r in resource_list if resource_name(r) is not None]
```
gregory-schiano, line 209: Put the related statements close together (`resource_list` and `resource` definition)
And the names are a bit ambiguous, try to be more specific (`new_resource` or `existing_resources`)

src/resource_manager/resource_manager.py
```diff
@@ -0,0 +1,250 @@
...
+                continue
+            if res_name == excluded_resource_name:
+                continue
+            self._delete_resource(name=res_name)
+            logger.info(
+                "%s deleted in namespace %s with name %s",
+                self._name,
+                self._namespace,
+                resource,
+            )
```
gregory-schiano, line 250: By filtering by name 2 deployments will silently compete, Weii suggested to add the relation ID or the app name in the resource name to distinguish

gregory-schiano: An alternative would be to check (if available) if there are any other deployment of the same charm

src/charm.py
```diff
@@ -26,11 +27,14 @@
...
-from gateway_definition import get_config
+from resource_definition import GatewayResourceDefinition, InvalidCharmConfigError
+from resource_manager.gateway import CreateGatewayError, GatewayResourceManager
+from resource_manager.resource_manager import InsufficientPermissionError, InvalidResourceError
 from tls_relation import TLSRelationService
 
 TLS_CERT = "certificates"
-LOGGER = logging.getLogger(__name__)
+logger = logging.getLogger(__name__)
+CREATED_BY_LABEL = "gateway-api-integrator.charm.juju.is/managed-by"
```
gregory-schiano, line 37: We might want to use the labels created by Juju directly

src/resource_manager/gateway.py
```diff
@@ -0,0 +1,238 @@
...
+            for gateway_class in gateway_classes
+            if gateway_class.metadata
+        )
+        if configured_gateway_class not in gateway_class_names:
+            logger.error(
+                "Configured gateway class %s not present on the cluster.", configured_gateway_class
+            )
+            raise CreateGatewayError(f"Gateway class {configured_gateway_class} not found.")
+
+        resource.spec["gatewayClassName"] = configured_gateway_class
```
gregory-schiano, line 119: It's more a charm config validation logic, it should be done in the `resource_definition.py`

src/resource_manager/gateway.py
```diff
@@ -0,0 +1,238 @@
...
+        """Generate a Gateway resource from a gateway resource definition.
+
+        Args:
+            definition: The gateway resoucre definition to use.
+
+        Returns:
+            A dictionary representing the gateway custom resource.
+        """
+        gateway = self._gateway_generic_resource(
+            apiVersion="gateway.networking.k8s.io/v1",
```
gregory-schiano, line 132: Use the `CUSTOM_GROUP_NAME` constant here

src/resource_manager/gateway.py
```diff
@@ -0,0 +1,238 @@
...
+            kind="Gateway",
```
gregory-schiano, line 133: You should use the `GATEWAY_RESOURCE_NAME` constant here

gregory-schiano (commented):
Reviewed with Thanh during an Architecture review call

requirements.txt
Thanhphan1147: PR for this : https://github.com/canonical/gateway-api-integrator-operator/pull/25

src/charm.py
Thanhphan1147: PR for this: https://github.com/canonical/gateway-api-integrator-operator/pull/21

PR #5, opened 2024-06-07
ISD-1986/ISD-1987 Refactor tls service and separate charm state

Thanhphan1147:
This PR changes how charm state is managed/initialized in the charm. It also changes how the charm do validation of required config as well as separate the `tls_relation` module into a charm state component (`TLSInformation`) and business logic component (the original `TLSRelationService`).

Validation of required config will be done using pydantic during initialization of each charm state component. For example, a missing config will cause `CharmConfig.from_charm` to raise an `InvalidCharmConfigError`. exceptions are caught and acted upon with a decorator `block_if_invalid_config_or_missing_integration` located at `state/vaildation.py`

### Checklist


Thanhphan1147:
@gregory-schiano can you please review this PR? 

Thanks!

src/state/config.py
```diff
@@ -0,0 +1,77 @@
...
+    def __init__(self, msg: str):
+        """Initialize a new instance of the InvalidCharmConfigError exception.
+
+        Args:
+            msg (str): Explanation of the error.
+        """
+        self.msg = msg
+
+
+@dataclasses.dataclass(frozen=True)
```
javierdelapuente, line 30: `class CharmConfig(BaseModel, frozen=True)` can be used too.

In the case of going through dataclasses, another option is to use pydantic dataclasses (see https://docs.pydantic.dev/latest/concepts/dataclasses/)

src/state/tls.py
```diff
@@ -0,0 +1,79 @@
...
+class TLSInformation:
+    """A component of charm state containing resource definition for kubernetes secret.
+
+    Attrs:
+        tls_requirer_integration: The integration instance with a TLS provider.
+        tls_certs: A dict of hostname: certificate obtained from the relation.
+        tls_keys: A dict of hostname: private_key stored in juju secrets.
+    """
+
+    tls_requirer_integration: Relation
```
javierdelapuente, line 42: This is a bit tricky, as one of the goals of charmstate is [`Code should not directly interact with the state provided by Juju`](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619).

In this case, an ops model (that is not inmutable) is in the state.

There is clearly a reason to have that information available (update the relation data, grant secrets...), but not sure if there could be better place to hide it.

javierdelapuente (commented):
Good job.! Just a few comments.

src/charm.py
```diff
@@ -106,20 +103,12 @@ def _reconcile(self) -> None:
             kubeconfig = KubeConfig.from_service_account()
             client = Client(config=kubeconfig, field_manager=self.app.name)
         except ConfigError as exc:
-            logger.exception("Error initializing the lightkube client.")
+            logger.error("Error initializing the lightkube client: %s", exc)
```
weiiwang01, line 106: I think `%r` might be better than `%s` to format exceptions, as `%s` will only keep the message of the exception, without the exception class.

src/state/config.py
```diff
@@ -0,0 +1,77 @@
...
+class CharmConfig(BaseModel):
+    """A component of charm state that contains the charm's configuration.
+
+    Attrs:
+        gateway_class (_type_): _description_
```
weiiwang01, line 35: Is the `_type_` and `_description_` here a placeholder?

weiiwang01 (dismissed):
LGTM, add some minor nitpicks

src/state/config.py
Thanhphan1147: Yes, I updated the relevant fields with proper description.

src/state/config.py
Thanhphan1147: thanks! I switched to pydantic dataclass as validation was not being done properly

src/charm.py
```diff
@@ -106,20 +103,12 @@ def _reconcile(self) -> None:
...
-            self.unit.status = BlockedStatus("Invalid charm configuration")
-            return
-
-        tls_certificates_relation = self._tls.get_tls_relation()
-        if not tls_certificates_relation:
-            self.unit.status = BlockedStatus("Waiting for TLS.")
-            return
+        config = CharmConfig.from_charm(self)
+        gateway_resource_definition = GatewayResourceDefinition.from_charm(self)
+        _ = TLSInformation.from_charm(self)
```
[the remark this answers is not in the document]
Thanhphan1147: Good point, I added a comment here.

src/state/tls.py
Thanhphan1147: I see your point, we can potentially exclude this from charm state and call every method of the `tls_certificate` module with an additional argument (the relation instance) which will solve the mutability question. I'll move the discussion internally and let's discuss there.

weiiwang01 (dismissed):

src/charm.py
```diff
@@ -111,25 +108,19 @@ def _reconcile(self) -> None:
...
-
-        tls_certificates_relation = self._tls.get_tls_relation()
-        if not tls_certificates_relation:
-            self.unit.status = BlockedStatus("Waiting for TLS.")
-            return
+        config = CharmConfig.from_charm(self)
+        gateway_resource_definition = GatewayResourceDefinition.from_charm(self)
+        # This line is currently here to validate the existence of the certificates relation.
+        # This charm state component will be used by the upcoming SecretResourceManager.
+        _ = TLSInformation.from_charm(self)
```
yanksyoon, line 115: Nitpick: Would `_` be required here? (feel free to resolve!)

src/charm.py
```diff
@@ -152,19 +143,17 @@ def _on_start(self, _: typing.Any) -> None:
...
         """
         hostname = event.params["hostname"]
-        tls_certificates_relation = self._tls.get_tls_relation()
-        if not tls_certificates_relation:
-            event.fail("Certificates relation not created.")
-            return
+        tls_information = TLSInformation.from_charm(self)
 
-        tls_rel_data = tls_certificates_relation.data[self.app]
+        tls_rel_data = tls_information.tls_requirer_integration.data[self.app]
```
yanksyoon, line 156: Just checking in: could this raise a `KeyError` if the self.app does not exist here?

yanksyoon: What I've done previously for integration data was to model them and initialize the DTO like so:
https://github.com/canonical/github-runner-operator/blob/main/src/charm_state.py#L771-L793
(would this be helpful here?)

src/resource_manager/resource_manager.py
```diff
@@ -136,11 +137,12 @@ def _name(self) -> str:
         """
 
     @abc.abstractmethod
-    def _gen_resource_from_definition(self, definition: ResourceDefinition) -> AnyResource:
+    def _gen_resource(self, definition: ResourceDefinition, config: CharmConfig) -> AnyResource:
         """Abstract method to generate a resource from ingress definition.
 
         Args:
             definition: Ingress definition to use for generating the resource.
+            config: The charm's configuration.
```
yanksyoon, line 145: I'm not so sure if we want to pass the whole charm config without modelling them for the module here :)

yanksyoon: I'm wondering if the CharmConfig class could be renamed to GatewayConfig to provide a little more suiting context for the module? :D

yanksyoon (dismissed):
LGTM, minor nitpicks and some questions :D Thank you!

src/resource_manager/resource_manager.py
Thanhphan1147: That's a valid point, but the issue here is that the `resource_manager` module is an abstract class, you'll notice that the `gen_resource`  method is not implemented, so I'm passing the entire config class here since even though a `gateway` resource manager might only need the hostname and gateway class, an `httpRoute` resource manager for example might need other things (like header rewriting configuration).

src/tls_relation.py
```diff
@@ -231,24 +197,7 @@ def certificate_expiring(  # type: ignore[no-untyped-def]
...
-            A dictionary indexed by domain, containing the decrypted private keys.
-        """
-        decrypted_private_keys = {
-            str(hostname): self._get_decrypted_key(
-                (private_key := self._get_private_key(hostname))["key"], private_key["password"]
-            )
-            for hostname in self.keys
-            if hostname
-        }
-        return decrypted_private_keys
```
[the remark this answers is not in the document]
Thanhphan1147: Will add this back in the next PR about managing secrets and https listener

src/state/tls.py
```diff
@@ -0,0 +1,79 @@
...
+        Args:
+            charm (ops.CharmBase): The gateway-api-integrator charm.
+
+        Raises:
+            TlsIntegrationMissingError: When integration is not ready.
+
+        Returns:
+            TLSInformation: Information about configured TLS certs.
+        """
+        tls_requirer_integration = charm.model.get_relation(TLS_CERTIFICATES_INTEGRATION)
```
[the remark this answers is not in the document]
Thanhphan1147: I tried things out a bit and the issue here is that when we store the relation object here (in charm state), the validation ensures that the relation exists and the relation object in charm state is of type `Relation` (rather than `Relation | Any` when doing `model.get_relation()`). Since the lib and our `tls_relation` module are designed in different ways (the lib doesn't assume that the relation always exist while our `tls_relation` module does the opposite and assumes that the relation exist). We'd have to significantly rewrite one or the other for them to work. I think we'd need to discuss on this further and if changes are required a follow-up PR would be more suited. I'm marking this as resolved for now. 

Thanks!

src/charm.py
Thanhphan1147: Thanks, this is a valid point, I mentioned it in the discussion started by Greg above, so I'm marking this as resolved as well.

src/charm.py
Thanhphan1147: I removed it to avoid any confusion

src/resource_manager/resource_manager.py
yanksyoon: I see, thanks for the explanation!

src/charm.py
```diff
@@ -282,50 +250,50 @@ def _certificate_revoked(self, revoke_list: typing.List[str]) -> None:  # noqa:
...
         Args:
             event: The event that fires this method.
         """
-        tls_certificates_relation = self._tls.get_tls_relation()
-        if not tls_certificates_relation:
+        try:
+            tls_information = TLSInformation.from_charm(self)
+        except (TlsIntegrationMissingError, InvalidCharmConfigError) as exc:
             self.unit.status = WaitingStatus("Waiting for certificates relation to be created")
+            logger.error(
```
yanksyoon, line 272: (Please feel free to implement on a followup PR), logger.exception can, by default, print exception stack trace. Hence there won't be a need to pass in exception (`logger.error("<message related to error> %s", exc)` -> `logger.exception("<message related to error>")`)

yanksyoon (approved):
LGTM! :rocket:

weiiwang01 (approved):

src/state/tls.py
Thanhphan1147: I'm marking this as resolved and move this discussion internally since this is linked to Greg's comment about using lib to access relation data

src/charm.py
Thanhphan1147: Thanks, this one might be an unintended change when I was keeping all the different PRs in sync, will change this in an upcoming PR.

PR #13, opened 2024-06-08
use square brackets instead of .get() for relation data

Thanhphan1147:
Discussed internally, since we check for the existence of the keys before returning the results, we use `[]` to access the value since we'll have an exception flow to the charm if somehow it fails. It's  a better approach than using `.get()` to return invalid data.

### Checklist


javierdelapuente (approved):

arturo-seijas (commented):
How would the cmd line output be like if this fails? I don't think it will be very informative to the user. Can these keys actually be absent?

Thanhphan1147:
I think it will only show the charm going into error state. Like I mentioned above this should never happen, but it's just that we've decided to use `[]` in case something really wrong `does` happens, it'd generate an exception and we'll know about it.

cbartz (approved):

PR #18, opened 2024-06-17
Add unit and integration tests

Thanhphan1147:
This PR adds unit tests to raise coverage to meet the 95% target and add integration tests to cover the main use-cases.

### Checklist


Thanhphan1147:
Hi @gregory-schiano , could you take a look at this pr and leave your review? Thanks!

tests/unit/test_certificates_integration.py
```diff
@@ -23,3 +30,191 @@ def test_generate_password(harness: Harness):
...
+    act: Initialize TLSInformation state component.
+    assert: TLSIntegrationMissingError is raised.
+    """
+    harness.begin()
+    with pytest.raises(TlsIntegrationMissingError):
+        TLSInformation.from_charm(harness.charm, harness.charm.certificates)
+
+
+@pytest.mark.usefixtures("client_with_mock_external")
+def test_cert_relation_certificate_expiring(
```
gregory-schiano, line 74: This test has no assertion

tests/unit/test_charm.py
```diff
@@ -0,0 +1,184 @@
...
+        "certificates", "self-signed-certificates", app_data=certificates_relation_data
+    )
+    harness.begin()
+
+    with pytest.raises(LightKubeInitializationError):
+        harness.update_config(config)
+
+
+@pytest.mark.usefixtures("patch_lightkube_client")
+def test_deploy_with_initial_hooks(harness: Harness):
```
gregory-schiano, line 78: There's no assertion in this test

gregory-schiano (commented):
Few comments
Also:
- The tests are globally really looking like functional tests or integration tests. It's ok for now, but we might want to have less tests that fully uses harness, and be more unittests
- I didn't see any integration tests despite what's stated in the PR titel

tests/unit/test_certificates_integration.py
Thanhphan1147: I added assertion to this test, thanks!

tests/unit/test_charm.py
Thanhphan1147: This test is redundant so I removed it.

Thanhphan1147:
@gregory-schiano As discussed internally I've added additional assertions in the main integration tests. Can you take a look again and leave your review? Thanks!

tests/integration/conftest.py
```diff
@@ -86,7 +89,48 @@ async def ingress_requirer_application_fixture(
...
+
+
+@pytest_asyncio.fixture(scope="module", name="any_charm_ingress_requirer")
+async def any_charm_ingress_requirer_fixture(model: Model):
+    """Deploy any-charm and patch it with ingress lib."""
+    any_app_name = "any-ingress"
+    ingress_lib_url = (
+        "https://github.com/canonical/nginx-ingress-integrator-operator"
+        "/raw/main/lib/charms/nginx_ingress_integrator/v0/ingress.py"
+    )
```
gregory-schiano, line 102: This is not the default location of the ingress library which is there:
`https://raw.githubusercontent.com/canonical/charm-relation-interfaces/main/lib/charms/interfaces/v2/ingress.py`

But isn't there a way to call fetch-lib instead ? I don't know if juju or pytest_operator provide this feature

tests/integration/test_charm.py
```diff
@@ -60,10 +50,37 @@ async def test_deploy(
...
     gateway = lightkube_client.get(gateway_generic_resource_class, name=application.name)
-    assert len(gateway.spec["listeners"]) == 2
-    secret: Secret = lightkube_client.get(
-        Secret, name=f"{application.name}-secret-{TEST_EXTERNAL_HOSTNAME_CONFIG}"
+    gateway_lb_ip = gateway.status["addresses"][0]["value"]  # type: ignore
+    assert gateway_lb_ip, "LB address not assigned to gateway"
+
+    ingress_url = (
+        f"http://{gateway_lb_ip}/{application.model.name}-{ingress_requirer_application.name}"
+    )
```
gregory-schiano, line 58: A better way to get that url is to add an action on AnyCharm that returns the URL provided in the relation, it allows you to test that the URL in the relation is also correct

gregory-schiano: And since you want HTTP you can do an URL parse and change the scheme

gregory-schiano (commented):
Few minor comments

tests/integration/test_charm.py
Thanhphan1147: I wanted to use an actual charm with workload for this, so I added a helper function to get the ingress url from the unit data. I also agree with adding an action on AnyCharm and I'll add that in a follow up PR

tests/integration/conftest.py
Thanhphan1147: I haven't seen `fetch-libs` calls in pytest_operator, I fixed the location of the library for now

gregory-schiano (approved):
LGTM

tests/integration/test_actions.py
```diff
@@ -0,0 +1,24 @@
...
+"""Integration test for actions."""
+
+import pytest
+from juju.application import Application
+
+from .conftest import TEST_EXTERNAL_HOSTNAME_CONFIG
+
+
+@pytest.mark.abort_on_fail
+async def test_deploy(
```
mthaddon, line 13: I feel like this should be named `test_get_certificate_action` or something similar rather than `test_deploy`. What do your think? Since this is such a small change it might be better to address in a separate PR (if you agree) to avoid blocking this from landing.

mthaddon (approved):

PR #28, opened 2024-06-19
ISD-1947 Implement secret resource manager and add HTTPS listener to gateway

Thanhphan1147:
Applicable spec: ISD-146 (internal)

### Overview

Similar to ingress, gateway uses kuberentes secret to bind a listener with the corresponding TLS certificate, we need to therefore manage this resource in our charm. 

This PR adds the following :
* update to tls_information, during initialization of this charm state component, we store the password used to decrypt the private key fetched from juju secrets alongside the private key itself. This password is used to decrypt the private key when generating the k8s resource.
* Add an additional parameter to integration test to allow verification of the created gateway and secret resource.
* Add a secretResourceDefinition class.
* Add a secretResourceManager class.
* Add HTTPS listener that uses the created k8s secret when creating the gateway resource.

### Checklist


Thanhphan1147:
Hi, @gregory-schiano can you take a look and review this PR? Thanks!

src/resource_manager/gateway.py
```diff
@@ -62,29 +57,40 @@ def _label_selector(self) -> str:
...
         """
+        tls_secret_name = f"{state.secret_resource_name_prefix}-{state.external_hostname}"
         gateway = self._gateway_generic_resource(
             apiVersion="gateway.networking.k8s.io/v1",
             kind="Gateway",
-            metadata=ObjectMeta(name=definition.gateway_name, labels=self._labels),
+            metadata=ObjectMeta(name=state.gateway_name, labels=self._labels),
             spec={
-                "gatewayClassName": config.gateway_class,
+                "gatewayClassName": state.gateway_class,
```
yanksyoon, line 78: Just out of curiosity, could this spec dict be typed?

src/resource_manager/gateway.py
```diff
@@ -140,7 +146,7 @@ def _delete_resource(self, name: str) -> None:
             name=name,
         )
 
-    def gateway_address(self, name: str) -> Optional[str]:
+    def gateway_address(self, name: str) -> typing.Optional[str]:  # pragma: no cover
```
yanksyoon, line 149: Would you be able to help explain the reason for the `no cover` as a comment if possible please? ;D

src/resource_manager/secret.py
```diff
@@ -0,0 +1,139 @@
...
+
+        Args:
+            labels: Label to be added to created resources.
+            client: Initialized lightkube client.
+        """
+        self._client = client
+        self._labels = labels
+
+    @property
+    def _name(self) -> str:
```
yanksyoon, line 60: Would you be able to, in the docstrings, perhaps mention that this is to fulfill the ResourceManager protocol please? :D

src/resource_manager/secret.py
```diff
@@ -0,0 +1,139 @@
...
+        )
+
+    @map_k8s_auth_exception
+    def _list_resource(self) -> typing.List[Secret]:
+        """List secret resources in a given namespace based on a label selector.
+
+        Returns:
+            A list of matched secret resources.
+        """
+        return list(self._client.list(res=Secret, labels=self._labels))
```
yanksyoon, line 130: Would `self._client.list(...)` be sufficient here?

src/state/base.py
```diff
@@ -0,0 +1,29 @@
...
+from .gateway import GatewayResourceDefinition
+from .secret import SecretResourceDefinition
+from .tls import TLSInformation
+
+Components = typing.TypeVar(
+    "Components", GatewayResourceDefinition, CharmConfig, SecretResourceDefinition, TLSInformation
+)
+
+
+class State:  # pylint: disable=too-few-public-methods
```
yanksyoon, line 18: Just out of curiosity, would this wrapper be safe enough to catch the errors statically? For instance, when we forget to pass in an argument, can it catch it statically? The reason I'm asking is because then the checks would need to delegate the responsibility to integration tests which take longer than lint/unit tests.

src/state/secret.py
```diff
@@ -0,0 +1,35 @@
...
+
+import ops
+
+
+@dataclasses.dataclass(frozen=True)
+class SecretResourceDefinition:
+    """A component of charm state that contains secret resource definition.
+
+    Attrs:
+        secret_resource_name_prefix (str):
```
yanksyoon, line 16: Perhaps the `(str)` is not required here :D

yanksyoon (dismissed):
LGTM! minor questions and nitpicks :)

src/resource_manager/gateway.py
Thanhphan1147: Good point, I think I looked into this when I started with the gateway implementation. I just looked again and it seems like there's no typed spec for `GenericNamespacedResource`

src/resource_manager/gateway.py
Thanhphan1147: Yes, I put `no cover` here since this code can waits for up to 60 seconds and it would slow down unit tests quite a bit if we test all the branches of this method

src/resource_manager/secret.py
Thanhphan1147: I have a PR coming that will completely remove this unused property so I'm resolving this comment for now.

src/state/secret.py
Thanhphan1147: I updated the docstring here, thanks!

Thanhphan1147:
hi @gregory-schiano, can you take another look at the PR? thanks!

src/resource_manager/secret.py
Thanhphan1147: I just checked and unfortunately similar to gateway `client.list` returns an iterator while we need a list to be able to perform logic

src/resource_manager/gateway.py
```diff
@@ -27,8 +26,26 @@
...
+    Attrs:
+        gateway_name: The gateway resource's name
+        external_hostname: The configured gateway hostname.
+        gateway_class: The configured gateway class.
+        secret_resource_name_prefix: Prefix of the secret resource name.
+    """
+
+    gateway_name: str
+    external_hostname: str
+    gateway_class: str
```
gregory-schiano, line 47: Should probably be `gateway_class_name`

src/charm.py
```diff
@@ -131,20 +134,25 @@ def _reconcile(self) -> None:
         config = CharmConfig.from_charm(self, client)
 
         gateway_resource_definition = GatewayResourceDefinition.from_charm(self)
-        # This line is currently here to validate the existence of the certificates relation.
-        # This charm state component will be used by the upcoming SecretResourceManager.
-        TLSInformation.from_charm(self)
+        tls_information = TLSInformation.from_charm(self)
+        secret_resource_definition = SecretResourceDefinition.from_charm(self)
```
gregory-schiano, line 138: Is it still a secret resource definition ? Since it doesn't contain anymore the data allowed to fully create a Secret Resource

I'm wondering if we shouldn't rename the `State` as `ResourceDefinition` and that one differently

src/resource_manager/gateway.py
```diff
@@ -27,8 +26,26 @@
 GATEWAY_PLURAL = "gateways"
 
 
-class CreateGatewayError(Exception):
-    """Represents an error when creating the gateway resource."""
+@dataclasses.dataclass
+class GatewayState(State):
```
gregory-schiano, line 30: As mentioned in another comment, we should probably rename this to `GatewayResourceDefinition`

src/resource_manager/gateway.py
```diff
@@ -62,29 +79,40 @@ def _label_selector(self) -> str:
         return ",".join(f"{k}={v}" for k, v in self._labels.items())
 
     @map_k8s_auth_exception
-    def _gen_resource(self, definition: GatewayResourceDefinition, config: CharmConfig) -> dict:
+    def _gen_resource(self, state: State) -> dict:
```
gregory-schiano, line 82: As mentioned in another comment, we should probably rename this to `ResourceDefinition` and `resource_definition`

src/resource_manager/gateway.py
gregory-schiano: Can't you mock `time.time()` in your unittest with a side effect where you'll be able to control the effective waiting time ?

src/resource_manager/secret.py
```diff
@@ -0,0 +1,167 @@
...
+        tls_certs: A dict of hostname: certificate obtained from the relation.
+        tls_keys: A dict of hostname: private_key stored in juju secrets.
+        secret_resource_name_prefix: Prefix of the secret resource name.
+    """
+
+    gateway_class: str
+    external_hostname: str
+    tls_requirer_integration: Relation
+    tls_certs: dict[str, str]
+    tls_keys: dict[str, dict[str, str]]
```
gregory-schiano, line 47: Is there a case where multiple certs and keys are linked to a give `external_hostname` ? If not, shouldn't we directly set the corresponding cert and key ?

src/resource_manager/secret.py
```diff
@@ -0,0 +1,167 @@
...
+
+    # There are multiple representation PKCS8 is the default supported by nginx controller
+    return decrypted_key.private_bytes(
+        encoding=serialization.Encoding.PEM,
+        format=serialization.PrivateFormat.PKCS8,
+        encryption_algorithm=serialization.NoEncryption(),
+    ).decode()
+
+
+class SecretResourceManager(ResourceManager[Secret]):
```
gregory-schiano, line 73: Should probably be named `TLSSecretResourceManager` as it creates specifically secrets with TLS information and not a generic secret

src/charm.py
```diff
@@ -131,20 +134,25 @@ def _reconcile(self) -> None:
...
         )
+        secret_resource_manager = SecretResourceManager(self._labels, client)
 
         try:
-            gateway = gateway_resource_manager.define_resource(gateway_resource_definition, config)
-        except (CreateGatewayError, InvalidResourceError) as exc:
-            logger.exception("Error creating the gateway resource.")
-            raise RuntimeError("Cannot create gateway.") from exc
+            secret = secret_resource_manager.define_resource(
+                State(secret_resource_definition, config, tls_information)
```
gregory-schiano, line 148: Couldn't you create directly a `SecretState` here ? This will avoid using the magic attribute creation in `State`

src/charm.py
```diff
@@ -131,20 +134,25 @@ def _reconcile(self) -> None:
...
+            )
+            gateway = gateway_resource_manager.define_resource(
+                State(gateway_resource_definition, config, secret_resource_definition)
```
gregory-schiano, line 151: Same as the previous comment for `secret`

src/resource_manager/gateway.py
```diff
@@ -62,16 +63,30 @@ def _label_selector(self) -> str:
...
+        if (
+            len(args) != 2
+            or not isinstance(args[0], CharmConfig)
+            or not isinstance(args[1], SecretResourceDefinition)
+        ):
+            raise CreateGatewayError("_gen_resource called with the wrong parameters.")
+
+        config: CharmConfig
+        secret: SecretResourceDefinition
+        config, secret = args
```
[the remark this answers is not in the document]
Thanhphan1147: This has been updated following the discussion we had internally, so I'm resolving this conversation

src/resource_manager/secret.py
```diff
@@ -0,0 +1,157 @@
...
+                "tls.crt": tls_information.tls_certs[config.external_hostname],
+                "tls.key": _get_decrypted_key(
+                    tls_information.tls_keys[config.external_hostname]["key"],
+                    tls_information.tls_keys[config.external_hostname]["password"],
+                ),
+            },
+            type="kubernetes.io/tls",
+        )
+
+        logger.info("Generated secret resource: %s", secret)
```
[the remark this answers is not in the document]
Thanhphan1147: thanks, I removed it.

src/charm.py
Thanhphan1147: That makes sense, I've also reworked the initialization logic for this class so that it requires exactly those 3 components as parameters.

src/resource_manager/gateway.py
Thanhphan1147: That makes sense, I've renamed the class to GatewayResourceDefinition, the component to GatewayResourceInformation in all relevant places.

src/resource_manager/gateway.py
Thanhphan1147: I renamed this to ResourceDefinition as it indeed makes more sense

src/charm.py
Thanhphan1147: Yep, I renamed this to `SecretResourceInformation` and the class used in resource_manager is renamed to `SecretResourceDefinition`

src/resource_manager/gateway.py
Thanhphan1147: I renamed it to `gateway_class_name` and also in all other relevant places

src/resource_manager/secret.py
Thanhphan1147: makes sense, I renamed it

src/state/base.py
Thanhphan1147: I also subclassed `State` (now `ResourceDefinition`) and override the `__init__` method to explicitly require the necessary components

src/resource_manager/gateway.py
Thanhphan1147: I didn't know that this was something you can do, I've remove `no cover` here and I'll add tests for this method in https://github.com/canonical/gateway-api-integrator-operator/pull/18

src/resource_manager/secret.py
Thanhphan1147: I left this implementation as-is when bringing it over from the `nginx-ingress-integrator` charm since we're also planning to support multiple hostnames in the future, with this implementation a hostname maps to one cert and one key, so there's no cases where multiple certs are linked to a hostname. I think it's ok to keep it like this, what do you think?

yanksyoon (approved):
Thanks for the explanation! LGTM :D

yhaliaw (approved):

src/resource_manager/secret.py
gregory-schiano: Few things to concider:
- The current implementation requests a TLS cert only for the `external_hostname`
- Ingress permits to do `path` routing or `subdomain` routing, which means in the case of `subdomain` routing and multiple relation, we'll have multiple certificates which are subdomains
- TLS relation allows to use SANs (Subject Alternative Names) which allows multiple domains per certificate
- A future `gateway-route` relation will likely allow the requirer to specify which domain they want to use, which means the charm will have to support multiple domains certificate

That said, these's no cases where a single domain has multiple certificates (or just temporarily when a cert is getting renewed).
This data structure associate an `external_hostname` to a list of cert et keys, which doesn't make sense

Thanhphan1147: After internal discussion It's decided that we'll rework the attributes in the `SecretResourceDefinition` class in the secret_resource_manager module to only contains a single TLS cert instead of passing the entire dictionary. Since I already have 2 approvals for this PR I'll mark this conversation as resolve and make a follow-up PR to address those changes.

PR #29, opened 2024-06-19
Validate gateway class in charm state

Thanhphan1147:
### Overview

We can validate that the configured gateway_class is valid as soon as the charm state is initialized by listing the gateway classes available on the cluster. 

This requires moving the lightkube client initialization to a helper `get_client` method.

### Checklist


src/charm.py
```diff
@@ -47,8 +51,21 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
 
         Args:
             args: Variable list of positional arguments passed to the parent constructor.
+
+        Raises:
+            RuntimeError: If initialization of the lightkube client fails
```
cbartz, line 56: I think using RuntimeError is too generic, could you create a custom error class (e.g. `LightKubeInitializationError`)

src/charm.py
```diff
@@ -174,72 +163,52 @@ def _on_get_certificate_action(self, event: ActionEvent) -> None:
 
         event.set_results(
             {
-                f"certificate-{hostname}": tls_rel_data[f"certificate-{hostname}"],
-                f"ca-{hostname}": tls_rel_data[f"ca-{hostname}"],
-                f"chain-{hostname}": tls_rel_data[f"chain-{hostname}"],
+                f"certificate-{hostname}": tls_rel_data.get(f"certificate-{hostname}"),
+                f"ca-{hostname}": tls_rel_data.get(f"ca-{hostname}"),
+                f"chain-{hostname}": tls_rel_data.get(f"chain-{hostname}"),
```
cbartz, line 168: Why the revert here?

src/state/config.py
```diff
@@ -0,0 +1,116 @@
...
+        gateway_class = typing.cast(str, charm.config.get("gateway-class"))
+        # my-py don't know that charm.client will always be initialized here
+        client: Client = charm.client  # type: ignore[attr-defined]
+        gateway_class_generic_resource = create_global_resource(
+            CUSTOM_RESOURCE_GROUP_NAME, "v1", GATEWAY_CLASS_RESOURCE_NAME, GATEWAY_CLASS_PLURAL
+        )
+        gateway_classes = tuple(client.list(gateway_class_generic_resource))
+        if not gateway_classes:
+            logger.error("No gateway class available on cluster.")
+            raise InvalidCharmConfigError("No gateway class available on cluster.")
```
cbartz, line 75: Is this really a config error? `gateway_classes` does not depend on the charm config.

src/state/tls.py
```diff
@@ -0,0 +1,79 @@
...
+
+    Attrs:
+        tls_requirer_integration: The integration instance with a TLS provider.
+        tls_certs: A dict of hostname: certificate obtained from the relation.
+        tls_keys: A dict of hostname: private_key stored in juju secrets.
+    """
+
+    tls_requirer_integration: Relation
+    tls_certs: Dict[str, str]
+    tls_keys: Dict[str, str]
```
cbartz, line 44:
```suggestion
    tls_certs: dict[str, str]
    tls_keys: dict[str, str]
```

cbartz (commented):
Looks good, some comments

src/state/config.py
Thanhphan1147: I was also not sure about this point but hesitated. I changed this to raise a more relevant error.

src/charm.py
Thanhphan1147: oops, this was unintended.

Thanhphan1147: fixed.

src/charm.py
Thanhphan1147: Indeed, I changed this to raise a custom error here.

Thanhphan1147:
Moving this back to `draft` as it's based on #5 and requires the PR to be merged first.

src/charm.py
```diff
@@ -41,6 +42,48 @@
...
+
+    Args:
+        field_manager (str): field manager for Server-side Apply
+        namespace (str): The k8s namespace in which resources are managed.
+
+    Raises:
+        LightKubeInitializationError: _description_
+
+    Returns:
+        Client: _description_
```
cbartz, line 57: Can you please explain `_description_`?

src/charm.py
```diff
@@ -110,13 +140,14 @@ def _reconcile(self) -> None:
             RuntimeError: when initializing the lightkube client fails,
             or when creating the gateway resource fails.
         """
-        config = CharmConfig.from_charm(self)
+        client = get_client(self.app.name, self.model.name)
+        config = CharmConfig.from_charm(self, client)
         gateway_resource_definition = GatewayResourceDefinition.from_charm(self)
         _ = TLSInformation.from_charm(self)
```
cbartz, line 146: Why is the call needed if the argument is not used? Are we relying on side effects here?

src/charm.py
```diff
@@ -163,17 +194,17 @@ def _on_get_certificate_action(self, event: ActionEvent) -> None:
...
+                f"chain-{hostname}": tls_rel_data.get[f"chain-{hostname}"],
             }
         )
 
     @block_if_invalid_config_or_missing_integration(defer=True)
     def _on_certificates_relation_created(self, _: RelationCreatedEvent) -> None:
         """Handle the TLS Certificate relation created event."""
         tls_information = TLSInformation.from_charm(self)
-        config = CharmConfig.from_charm(self)
+        config = CharmConfig.from_charm(self, get_client(self.app.name, self.model.name))
```
cbartz, line 207: Minor: I prefer to have a separate variable for method calls and not include them as arguments, as this makes debugging easier.

src/state/tls.py
```diff
@@ -0,0 +1,78 @@
...
+        tls_keys: A dict of hostname: private_key stored in juju secrets.
+    """
+
+    tls_requirer_integration: Relation
+    tls_certs: dict[str, str]
+    tls_keys: dict[str, str]
+
+    @classmethod
+    def from_charm(cls, charm: ops.CharmBase) -> "TLSInformation":
+        """Create a resource definition from charm instance.
```
cbartz, line 47: I think this line is not correct as you are not creating a resource definition here.

src/charm.py
Thanhphan1147: Yes, we use the `validate_config_and_integration` decorator to catch errors raised by `TLSInformation.from_charm` if the tls relation is not ready and set the charm to blocked state when that happens. It is defined [here](https://github.com/canonical/gateway-api-integrator-operator/blob/1c46079ae141ee3f891cac03a37ca2994a372289/src/state/validation.py#L21)

src/charm.py
Thanhphan1147: Makes sense, I created separate variables in relevant places. It also makes it clearer that `self.app.name` is used as field_manager and `self.model.name` as k8s namespace.

src/state/tls.py
Thanhphan1147: Fixed the docstring. Thanks!

src/charm.py
Thanhphan1147: Fixed the docstring. Thanks!

src/state/config.py
```diff
@@ -42,21 +67,51 @@ class CharmConfig:
...
+    def from_charm(cls, charm: ops.CharmBase, client: Client) -> "CharmConfig":
         """Create a CharmConfig class from a charm instance.
 
         Args:
             charm (ops.CharmBase): The gateway-api-integrator charm.
+            client (lightkube.Client): The lightkube client
 
         Raises:
-            InvalidCharmConfigError: When validation of the charm's config failed.
+            InvalidCharmConfigError: _description_
```
cbartz, line 78: `_description_` needs to be replaced

src/charm.py
```diff
@@ -41,6 +42,48 @@
...
+        msg (str): Explanation of the error.
+    """
+
+    def __init__(self, msg: str):
+        """Initialize a new instance of the InvalidCharmConfigError exception.
+
+        Args:
+            msg (str): Explanation of the error.
+        """
+        self.msg = msg
```
cbartz, line 84: I do not really see the benefit of this attribute as you can usually get the same effect using `str(e)`. Does also apply to other exceptions defined in this PR.

src/charm.py
```diff
@@ -173,7 +209,10 @@ def _on_get_certificate_action(self, event: ActionEvent) -> None:
     def _on_certificates_relation_created(self, _: RelationCreatedEvent) -> None:
         """Handle the TLS Certificate relation created event."""
         tls_information = TLSInformation.from_charm(self)
-        config = CharmConfig.from_charm(self)
+
+        field_manager = self.app.name
+        namespace = self.model.name
+        config = CharmConfig.from_charm(self, get_client(field_manager, namespace))
```
cbartz, line 215: Minor: create a separate variable for `get_client` (also applies to line 228 and 337)

cbartz (commented):
Minor comments

src/charm.py
cbartz: Hm, is there another way to achieve the same goal in a more explicit way to improve code readability? It seems you must have quite some internal knowledge of the code base to understand this line.

src/charm.py
Thanhphan1147: Fixed!

src/state/config.py
Thanhphan1147: Fixed, thanks!

src/charm.py
Thanhphan1147: I was thinking about this as well but did it this was as most of our charms implement errors like this. I removed this in relevant classes.

src/charm.py
```diff
@@ -173,7 +198,11 @@ def _on_get_certificate_action(self, event: ActionEvent) -> None:
     def _on_certificates_relation_created(self, _: RelationCreatedEvent) -> None:
         """Handle the TLS Certificate relation created event."""
         tls_information = TLSInformation.from_charm(self)
-        config = CharmConfig.from_charm(self)
+
+        field_manager = self.app.name
+        namespace = self.model.name
+        client = get_client(field_manager, namespace)
```
javierdelapuente, line 204: I think I am not seeing something obvious, but why not instantiate this variable as a class variable in the `__init__` method?

javierdelapuente (approved):

src/charm.py
```diff
@@ -41,6 +43,36 @@
 CREATED_BY_LABEL = "gateway-api-integrator.charm.juju.is/managed-by"
 
 
+@lru_cache(maxsize=1)
+def get_client(field_manager: str, namespace: str) -> Client:
+    """Initialize the lightkube client with the correct namespace and field_manager.
+
+    Args:
+        field_manager (str): field manager for server side apply when patching resources.
+        namespace (str): The k8s namespace in which resources are managed.
```
yanksyoon, line 52: Just out of curiosity, may I know why the type hint in docstrings? I've had feedbacks before that since we already have type hints to the arguments, we don't quite need to repeat it in the docstrings :D

src/resource_manager/decorator.py
yanksyoon, line 1: Would it help to call this something within context such as `permission.py`? I think the name `decorator` might be a bit too generic - as something like `utils.py` :D

src/state/gateway.py
```diff
@@ -13,21 +13,20 @@ class GatewayResourceDefinition:
     """Base class containing kubernetes resource definition.
 
     Attrs:
-        namespace: The gateway resource's namespace.
         gateway_name: The gateway resource's name
     """
 
-    namespace: str
+    # We're expecting more fields to be added.
```
yanksyoon, line 19: Would this indicate that we're ignoring a lint error somewhere? If so, I think this could go right above the line in which we're doing the `ignore` :)

src/state/gateway.py
```diff
@@ -13,21 +13,20 @@ class GatewayResourceDefinition:
...
     gateway_name: str
 
     @classmethod
     def from_charm(cls, charm: ops.CharmBase) -> "GatewayResourceDefinition":
         """Create a resource definition from charm instance.
 
         Args:
-            charm (ops.CharmBase): _description_
+            charm (ops.CharmBase): The charm instance.
```
yanksyoon, line 27:
```suggestion
            charm: The charm instance.
```
Nitpick: We can remove the type hint in the docstring :D

src/state/tls.py
```diff
@@ -14,19 +13,7 @@
...
-    """
-
-    def __init__(self, msg: str):
-        """Initialize a new instance of the TlsIntegrationMissingError exception.
-
-        Args:
-            msg (str): Explanation of the error.
-        """
-        self.msg = msg
+    """Exception raised when _situation_."""
```
yanksyoon, line 16:
```suggestion
    """Exception raised when TLS integration has not yet been setup."""
```

src/state/validation.py
```diff
@@ -60,13 +61,15 @@ def wrapper(instance: C, *args: typing.Any) -> None:
...
+                state.config.GatewayClassUnavailableError,
+                InsufficientPermissionError,
             ) as exc:
                 if defer:
                     event: ops.EventBase
                     event, *_ = args
                     event.defer()
-                logger.exception("Wrong Charm Configuration")
-                instance.unit.status = ops.BlockedStatus(exc.msg)
+                logger.exception(str(exc))
```
yanksyoon, line 71:
```suggestion
                logger.exception("Invalid user configuration exception.")
```
I think since logger.exception already prints the exception stack trace, re-printing it would be a bit too verbose :D

yanksyoon (approved):
Minor nitpicks, thank you!

Thanhphan1147:
@gregory-schiano please review this PR, thanks!

src/state/gateway.py
gregory-schiano: No I asked thanh to add that comment because using Pydantic for a such simple data object is a bit overkill. Adding that comment let know the reviewer that this class will be extended in a coming PR

src/state/validation.py
```diff
@@ -60,13 +61,15 @@ def wrapper(instance: C, *args: typing.Any) -> None:
             except (
                 state.config.InvalidCharmConfigError,
                 state.tls.TlsIntegrationMissingError,
+                state.config.GatewayClassUnavailableError,
+                InsufficientPermissionError,
```
gregory-schiano, line 65: Might be interesting now to have a parent exception class to simplify this

src/charm.py
javierdelapuente: Maybe then another option is to have a `client` property in the Charm (and cached_property if that is necessary)

src/charm.py
Thanhphan1147: It came with the doc string generator tool that I used, since I don't the the downside of having type hints there I didn't remove them. I'll remove the type hints to be consistent with other charms.

src/resource_manager/decorator.py
Thanhphan1147: Makes sense, will rename it to `permission.py`

src/charm.py
```diff
@@ -41,6 +43,36 @@
 CREATED_BY_LABEL = "gateway-api-integrator.charm.juju.is/managed-by"
 
 
+@lru_cache(maxsize=1)
```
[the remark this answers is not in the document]
Thanhphan1147: Makes sense, I was putting it here in case of deferred events and just to be sure that the client is not initialized multiple times, I'll remove the cache decorator.

src/charm.py
Thanhphan1147: That makes sense as well, I'll create a follow-up PR and let's move this discussion there.

src/state/validation.py
Thanhphan1147: Yes, I added this in a future PR.

src/charm.py
```diff
@@ -41,6 +43,36 @@
...
+def get_client(field_manager: str, namespace: str) -> Client:
```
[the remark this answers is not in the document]
Thanhphan1147: makes sense, I will add this to a follow-up PR

src/state/validation.py
Thanhphan1147: thanks, will add this in a follow-up PR

src/charm.py
```diff
@@ -97,18 +129,11 @@ def _reconcile(self) -> None:
...
-            # Keep this consistent across client initializations
-            kubeconfig = KubeConfig.from_service_account()
-            client = Client(
-                config=kubeconfig, field_manager=self.app.name, namespace=self.model.name
-            )
-        except ConfigError as exc:
-            logger.exception("Error initializing the lightkube client.")
-            raise RuntimeError("Error initializing the lightkube client.") from exc
+        field_manager = self.app.name
+        namespace = self.model.name
```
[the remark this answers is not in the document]
Thanhphan1147: I did that initially but @cbartz suggested that I create local variables to be more explicit, I think I'll remove those 2 variables and use kw_args to stay explicit but remove the need for local variables.

PR #33, opened 2024-06-26
Implement ingress integration

Thanhphan1147:
Applicable spec: <link>

### Overview
Adds the following:
- Ingress integration and ingress provider lib
- HTTP_route resource definition from ingress provider lib
- Service resource manager
- http_route resource manager
- Publish ingress endpoint

### Checklist


src/charm.py
```diff
@@ -27,17 +32,24 @@
...
 from state.gateway import GatewayResourceInformation
+from state.http_route import HTTPRouteResourceInformation
 from state.tls import TLSInformation
 from state.validation import validate_config_and_integration
 from tls_relation import TLSRelationService, get_hostname_from_cert
 
 TLS_CERT = "certificates"
 logger = logging.getLogger(__name__)
 CREATED_BY_LABEL = "gateway-api-integrator.charm.juju.is/managed-by"
+INGRESS_RELATION = "gateway"
```
gregory-schiano, line 52: For coherence shouldn't `TLS_CERT` be named `TLS_CERT_RELATION` and both declared next to each other ?

src/charm.py
```diff
@@ -110,20 +123,19 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
     @validate_config_and_integration(defer=False)
-    def _reconcile(self) -> None:
-        """Reconcile charm status based on configuration and integrations.
-
-        Raises:
-            RuntimeError: when initializing the lightkube client fails,
-            or when creating the gateway resource fails.
-        """
+    def _reconcile(self) -> None:  # pylint: disable=too-many-locals
+        """Reconcile charm status based on configuration and integrations."""
```
gregory-schiano, line 136: I think the description should be more explicit and describe the steps executed in this method

src/resource_manager/http_route.py
```diff
@@ -0,0 +1,213 @@
...
+            spec["rules"] = [
+                {
+                    "filters": [
+                        {
+                            "type": "RequestRedirect",
+                            "requestRedirect": {"scheme": "https", "statusCode": 301},
+                        }
+                    ]
+                }
+            ]
```
gregory-schiano, line 151: Not sure this should be managed here, it's not a normal route but a "redirect" route. I'd probably create another resource manager named `HTTPtoHTTPSResourceManager` that inherits from `HTTPRouteResourceManager` and overrides _gen_resource

src/state/http_route.py
```diff
@@ -0,0 +1,77 @@
...
+        Returns:
+            HTTPRouteResourceInformation: Information about configured TLS certs.
+        """
+        ingress_integration = charm.model.get_relation(INGRESS_RELATION)
+        if ingress_integration is None:
+            raise IngressIntegrationMissingError("Ingress integration not ready.")
+        try:
+            integration_data = ingress_provider.get_data(ingress_integration)
+            application_name = integration_data.app.name
+            service_name = f"{application_name}-gateway-service"
```
weiiwang01, line 64: I think it might be better to use the gateway app name (`charm.app.name`), as an ingress requirer app can integrate with multiple gateway charm apps. In such cases, using the remote app name could lead to collisions.

src/resource_manager/http_route.py
```diff
@@ -0,0 +1,213 @@
...
+from .resource_manager import ResourceManager
+
+logger = logging.getLogger(__name__)
+
+CUSTOM_RESOURCE_GROUP_NAME = "gateway.networking.k8s.io"
+HTTP_ROUTE_RESOURCE_NAME = "HTTPRoute"
+HTTP_ROUTE_PLURAL = "httproutes"
+
+
+class HTTPRouteType(Enum):
```
weiiwang01, line 31: I think if this is defined as a `StrEnum`, there is no need to use `.value` in the subsequent code.

cbartz (commented):
Looks good, some comments.

src/resource_manager/http_route.py
Thanhphan1147: Thanks, I didn't know about this, I switched to using `StrEnum`

src/state/http_route.py
Thanhphan1147: I might not be understanding this correctly, `application_name` is used here to construct the path of the ingress endpoint similar to the `traefik-k8s` charm and if we plan to support multiple requirer charms then it's normal for the requirer charm to have an unique path. 

in case of multiple gateway charms each will have their unique gateway IP address so there shouldn't be a risk of collision.

In case where we would want to also handle cross-model integrations we should also prefix the path with the requirer's model name (not sure if this is supported yet in the charm relation interface.)

src/charm.py
Thanhphan1147: I renamed it and reordered the constants

src/resource_manager/http_route.py
```diff
@@ -0,0 +1,215 @@
...
+class HTTPRouteResourceDefinition(ResourceDefinition):
+    """A part of charm state with information required to manage gateway resource.
+
+    It consists of 2 components:
+        - HTTPRouteResourceInformation
+        - GatewayResourceDefinition
+
+    Attributes:
+        application_name: The requirer application name.
+        gateway_name: The gateway resource's name
```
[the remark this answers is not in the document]
Thanhphan1147: thanks, added the missing .

tests/integration/test_charm.py
```diff
@@ -39,6 +40,16 @@ async def test_deploy(
...
     )
+
+    await application.model.add_relation(
+        application.name, f"{ingress_requirer_application.name}:ingress"
+    )
+    await application.model.wait_for_idle(
+        apps=[ingress_requirer_application.name],
+        idle_period=30,
+        status="active",
+    )
```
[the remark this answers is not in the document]
Thanhphan1147: yes, they will be added in the next PR: https://github.com/canonical/gateway-api-integrator-operator/pull/18

src/resource_manager/http_route.py
Thanhphan1147: makes sense, I created a `HTTPRouteRedirectResourceManager` class

cbartz (approved):

src/state/http_route.py
weiiwang01: Here's a revised version of the text with improved English:

I was thinking about this line: https://github.com/canonical/gateway-api-integrator-operator/blob/ea8eacc7cdf6e5b7aa3880cc97b534c4a963e073/src/resource_manager/service.py#L87. 

This attribute is used as the name for service resources. If I remember correctly, there can't be two resources with the same name in the same namespace in Kubernetes?

gregory-schiano (approved):

src/state/http_route.py
Thanhphan1147: Ah, I see, this does indeed need to change, I'll make the necessary changes in a follow-up PR. Thanks!

src/charm.py
```diff
@@ -110,20 +123,19 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
 
     @validate_config_and_integration(defer=False)
-    def _reconcile(self) -> None:
-        """Reconcile charm status based on configuration and integrations.
-
-        Raises:
-            RuntimeError: when initializing the lightkube client fails,
-            or when creating the gateway resource fails.
-        """
+    def _reconcile(self) -> None:  # pylint: disable=too-many-locals
```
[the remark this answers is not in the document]
Thanhphan1147: thanks, I moved the method.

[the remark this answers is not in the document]
Thanhphan1147: I refactored this method completely and removed the pylint comment

src/charm.py
Thanhphan1147: I refactored this method to be more straigh-forward to understand and also updated the docstring to include more information

PR #50, opened 2024-07-12
ISD-2100 Refactor tls_relation module and charm state logic

Thanhphan1147:
Follow up to comments on #28 and internal discussions. The `tls_relation` module needs to be simplified and the `certificates` integration handling logic needs to be improved.

This PR refactors and tries to simplify the `tls_relation` module, the `tls_information` charm state component and the `SecretResourceDefinition` used to manage k8s secrets:

* `tls_relation` module is now initialized with the certificates requirer v3 lib, replacing all direct modification of the integration data with library calls. Remove useless code that stores the certificates information on the application's integration data.
* `TLSInformation` state component now no longer stores the tls integration integration.
* `SecretResourceDefinition` now is built directly from TLSInformation using `SecretResourceDefinition.from_tls_information` class method
* Simplify the `certificates` integration handling logic.
* Update tests to accomodate the above changes

### Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -72,8 +72,6 @@ class LightKubeInitializationError(Exception):
 class GatewayAPICharm(CharmBase):
     """The main charm class for the gateway-api-integrator charm."""
 
-    _authed = False
-
     def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
         """Init method for the class.
 
@@ -83,7 +81,7 @@ class GatewayAPICharm(CharmBase):
         super().__init__(*args)
 
         self.certificates = TLSCertificatesRequiresV3(self, TLS_CERT)
-        self._tls = TLSRelationService(self.certificates)
+        self._tls = TLSRelationService(self.model, self.certificates)
 
         self.framework.observe(self.on.config_changed, self._on_config_changed)
         self.framework.observe(self.on.start, self._on_start)
@@ -173,7 +171,7 @@ class GatewayAPICharm(CharmBase):
             event: Juju event
         """
         hostname = event.params["hostname"]
-        TLSInformation.from_charm(self, self.certificates)
+        TLSInformation.validate(self)
 
         for cert in self.certificates.get_provider_certificates():
             if get_hostname_from_cert(cert.certificate) == hostname:
@@ -191,7 +189,7 @@ class GatewayAPICharm(CharmBase):
     @validate_config_and_integration(defer=True)
     def _on_certificates_relation_created(self, _: RelationCreatedEvent) -> None:
         """Handle the TLS Certificate relation created event."""
-        TLSInformation.from_charm(self, self.certificates)
+        TLSInformation.validate(self)
         client = _get_client(field_manager=self.app.name, namespace=self.model.name)
         config = CharmConfig.from_charm(self, client)
         self._tls.certificate_relation_created(config.external_hostname)
@@ -199,7 +197,7 @@ class GatewayAPICharm(CharmBase):
     @validate_config_and_integration(defer=True)
     def _on_certificates_relation_joined(self, _: RelationJoinedEvent) -> None:
         """Handle the TLS Certificate relation joined event."""
-        TLSInformation.from_charm(self, self.certificates)
+        TLSInformation.validate(self)
         client = _get_client(field_manager=self.app.name, namespace=self.model.name)
         config = CharmConfig.from_charm(self, client)
         self._tls.certificate_relation_joined(config.external_hostname)
@@ -220,7 +218,7 @@ class GatewayAPICharm(CharmBase):
         Args:
             event: The event that fires this method.
         """
-        TLSInformation.from_charm(self, self.certificates)
+        TLSInformation.validate(self)
         self._tls.certificate_expiring(event)
 
     @validate_config_and_integration(defer=True)
@@ -230,7 +228,7 @@ class GatewayAPICharm(CharmBase):
         Args:
             event: The event that fires this method.
         """
-        TLSInformation.from_charm(self, self.certificates)
+        TLSInformation.validate(self)
         if event.reason == "revoked":
             self._tls.certificate_invalidated(event)
         if event.reason == "expired":
@@ -240,7 +238,7 @@ class GatewayAPICharm(CharmBase):
     @validate_config_and_integration(defer=True)
     def _on_all_certificates_invalidated(self, _: AllCertificatesInvalidatedEvent) -> None:
         """Handle the TLS Certificate relation broken event."""
-        TLSInformation.from_charm(self, self.certificates)
+        TLSInformation.validate(self)
         client = _get_client(field_manager=self.app.name, namespace=self.model.name)
         config = CharmConfig.from_charm(self, client)
         hostname = config.external_hostname
--- a/src/charm.py
+++ b/src/charm.py
@@ -16,27 +16,24 @@ from charms.tls_certificates_interface.v3.tls_certificates import (
 )
 from lightkube import Client, KubeConfig
 from lightkube.core.exceptions import ConfigError
-from ops.charm import ActionEvent, CharmBase, RelationCreatedEvent, RelationJoinedEvent
-from ops.jujuversion import JujuVersion
-from ops.main import main
-from ops.model import (
-    ActiveStatus,
-    BlockedStatus,
-    MaintenanceStatus,
-    Relation,
-    SecretNotFoundError,
-    WaitingStatus,
+from ops.charm import (
+    ActionEvent,
+    CharmBase,
+    RelationBrokenEvent,
+    RelationCreatedEvent,
+    RelationJoinedEvent,
 )
+from ops.main import main
+from ops.model import ActiveStatus, MaintenanceStatus, SecretNotFoundError, WaitingStatus
 
 from resource_manager.gateway import GatewayResourceDefinition, GatewayResourceManager
-from resource_manager.permission import InsufficientPermissionError
 from resource_manager.resource_manager import InvalidResourceError
 from resource_manager.secret import SecretResourceDefinition, TLSSecretResourceManager
-from state.config import CharmConfig, InvalidCharmConfigError
+from state.config import CharmConfig
 from state.gateway import GatewayResourceInformation
-from state.tls import TLSInformation, TlsIntegrationMissingError
+from state.tls import TLSInformation
 from state.validation import validate_config_and_integration
-from tls_relation import TLSRelationService
+from tls_relation import TLSRelationService, get_hostname_from_cert
 
 TLS_CERT = "certificates"
 logger = logging.getLogger(__name__)
@@ -86,7 +83,7 @@ class GatewayAPICharm(CharmBase):
         super().__init__(*args)
 
         self.certificates = TLSCertificatesRequiresV3(self, TLS_CERT)
-        self._tls = TLSRelationService(self.model)
+        self._tls = TLSRelationService(self.certificates)
 
         self.framework.observe(self.on.config_changed, self._on_config_changed)
         self.framework.observe(self.on.start, self._on_start)
@@ -131,7 +128,7 @@ class GatewayAPICharm(CharmBase):
         client = _get_client(field_manager=self.app.name, namespace=self.model.name)
         config = CharmConfig.from_charm(self, client)
         gateway_resource_information = GatewayResourceInformation.from_charm(self)
-        tls_information = TLSInformation.from_charm(self)
+        tls_information = TLSInformation.from_charm(self, self.certificates)
 
         gateway_resource_manager = GatewayResourceManager(
             labels=self._labels,
@@ -141,7 +138,9 @@ class GatewayAPICharm(CharmBase):
 
         try:
             secret = secret_resource_manager.define_resource(
-                SecretResourceDefinition(gateway_resource_information, config, tls_information)
+                SecretResourceDefinition.from_tls_information(
+                    tls_information, config.external_hostname
+                )
             )
             gateway = gateway_resource_manager.define_resource(
                 GatewayResourceDefinition(gateway_resource_information, config, tls_information)
@@ -149,9 +148,6 @@ class GatewayAPICharm(CharmBase):
         except InvalidResourceError as exc:
             logger.exception("Error creating resource")
             raise RuntimeError("Error creating resource.") from exc
-        except InsufficientPermissionError as exc:
-            self.unit.status = BlockedStatus(str(exc))
-            return
 
         self.unit.status = WaitingStatus("Waiting for gateway address")
         if gateway_address := gateway_resource_manager.gateway_address(gateway.metadata.name):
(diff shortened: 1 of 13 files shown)


Thanhphan1147:
Hi @gregory-schiano, can you please take a look at this PR and leave your review? Thanks!

src/charm.py
```diff
@@ -177,116 +173,55 @@ def _on_get_certificate_action(self, event: ActionEvent) -> None:
...
-            return
-
-        event.set_results(
-            {
-                f"certificate-{hostname}": tls_rel_data[f"certificate-{hostname}"],
-                f"ca-{hostname}": tls_rel_data[f"ca-{hostname}"],
-                f"chain-{hostname}": tls_rel_data[f"chain-{hostname}"],
-            }
-        )
+        TLSInformation.from_charm(self, self.certificates)
```
cbartz, line 176: This line is not obvious, I assume there are relevant side-effects in it, but I think there should at least be a comment pointing this out (because no return value is used and the name of the function does not explain the side-effects). It would be better if we could have a solution without relying on side-effects, as it is easier to reason about.

src/charm.py
```diff
@@ -177,116 +173,55 @@ def _on_get_certificate_action(self, event: ActionEvent) -> None:
...
+                return
+
+        event.fail(f"Missing or incomplete certificate data for {hostname}")
 
     @validate_config_and_integration(defer=True)
     def _on_certificates_relation_created(self, _: RelationCreatedEvent) -> None:
         """Handle the TLS Certificate relation created event."""
-        tls_information = TLSInformation.from_charm(self)
-
+        TLSInformation.from_charm(self, self.certificates)
```
cbartz, line 194: same here

src/charm.py
```diff
@@ -177,116 +173,55 @@ def _on_get_certificate_action(self, event: ActionEvent) -> None:
...
-            config.external_hostname, tls_information.tls_requirer_integration
-        )
+        self._tls.certificate_relation_created(config.external_hostname)
 
     @validate_config_and_integration(defer=True)
     def _on_certificates_relation_joined(self, _: RelationJoinedEvent) -> None:
         """Handle the TLS Certificate relation joined event."""
-        tls_information = TLSInformation.from_charm(self)
-
+        TLSInformation.from_charm(self, self.certificates)
```
cbartz, line 202: same here

src/charm.py
```diff
@@ -177,116 +173,55 @@ def _on_get_certificate_action(self, event: ActionEvent) -> None:
...
-                self._tls.pop_relation_data_fields(
-                    [f"key-{hostname}", f"password-{hostname}"],
-                    tls_requirer_integration,  # type: ignore[arg-type]
-                )
-            except KeyError:
-                logger.warning("Relation data for %s already does not exist", hostname)
-            self.certificates.request_certificate_revocation(
-                certificate_signing_request=old_csr.encode()
-            )
+        TLSInformation.from_charm(self, self.certificates)
```
cbartz, line 223: same here

src/charm.py
```diff
@@ -295,41 +230,26 @@ def _on_certificate_invalidated(self, event: CertificateInvalidatedEvent) -> Non
...
-            tls_information = TLSInformation.from_charm(self)
-        except (TlsIntegrationMissingError, InvalidCharmConfigError) as exc:
-            self.unit.status = WaitingStatus("Waiting for certificates relation to be created")
-            logger.error(
-                "(%s) charm is not ready to handle this event, deferring: %s.", event, exc
-            )
-            event.defer()
-            return
-
+        TLSInformation.from_charm(self, self.certificates)
```
cbartz, line 233: same here

src/charm.py
```diff
@@ -295,41 +230,26 @@ def _on_certificate_invalidated(self, event: CertificateInvalidatedEvent) -> Non
...
+            self._tls.certificate_expiring(event)
         self.unit.status = MaintenanceStatus("Waiting for new certificate")
 
     @validate_config_and_integration(defer=True)
     def _on_all_certificates_invalidated(self, _: AllCertificatesInvalidatedEvent) -> None:
         """Handle the TLS Certificate relation broken event."""
-        tls_information = TLSInformation.from_charm(self)
-        tls_information.tls_requirer_integration.data[self.app].clear()
-
+        TLSInformation.from_charm(self, self.certificates)
```
cbartz, line 243: same here

src/resource_manager/gateway.py
```diff
@@ -92,38 +92,39 @@ def _label_selector(self) -> str:
         return ",".join(f"{k}={v}" for k, v in self._labels.items())
 
     @map_k8s_auth_exception
-    def _gen_resource(self, state: ResourceDefinition) -> dict:
+    def _gen_resource(self, resource_definition: ResourceDefinition) -> dict:
         """Generate a Gateway resource from a gateway resource definition.
 
         Args:
-            state: Part of charm state.
+            resource_definition: Part of charm state.
```
cbartz, line 99: Can you rephrase this description?

src/resource_manager/resource_manager.py
```diff
@@ -45,11 +45,11 @@ class ResourceManager(typing.Protocol[AnyResource]):
     """Abstract base class for a generic Kubernetes resource controller."""
 
     @abc.abstractmethod
-    def _gen_resource(self, state: ResourceDefinition) -> AnyResource:
+    def _gen_resource(self, resource_definition: ResourceDefinition) -> AnyResource:
         """Abstract method to generate a resource from ingress definition.
 
         Args:
-            state: Part of charm state consisting of one or several components.
+            resource_definition: Part of charm state consisting of one or several components.
```
cbartz, line 52: Can you rephrase this description?

src/resource_manager/secret.py
```diff
@@ -25,43 +23,54 @@
...
+            hostname: The requested hostname.
+
+        Raises:
+            CertificateDataNotReadyError: When the certificate data is not ready.
+
+        Returns:
+            SecretResourceDefinition: Information about the certificate.
         """
-        super().__init__(gateway_resource_information, charm_config, tls_information)
+        if hostname not in tls_information.tls_certs.keys():
```
cbartz, line 64:
```suggestion
        if hostname not in tls_information.tls_certs:
```

if its a dict , using `in` without `keys()` should suffice

tests/unit/conftest.py
```diff
@@ -95,3 +95,13 @@ def mock_certificate_fixture() -> str:
         "4+3+0/6Ba2Zlt9fu4PixG+XukQnBIxtIMjWp7q7xWp8F4aOW"
         "-----END CERTIFICATE-----"
     )
+    provider_cert_mock = MagicMock()
```
cbartz, line 98: consider using the `spec` attribute for MagicMock

cbartz (commented):
Thanks for the refactoring! I think you forgot to do src-docs. I think there are also missing unit tests, but I guess you are preparing a PR for those?

src/charm.py
gregory-schiano: True, I missed that one

src/state/tls.py
```diff
@@ -22,29 +30,36 @@ class TLSInformation:
...
             TlsIntegrationMissingError: When integration is not ready.
+            SecretNotSupportedError: When running with incompatible juju version.
 
         Returns:
             TLSInformation: Information about configured TLS certs.
         """
+        if not JujuVersion.from_environ().has_secrets:
+            raise SecretNotSupportedError(
+                "The charm requires the 'secrets' feature to be supported (juju version >= 3.0.3)."
+            )
```
gregory-schiano, line 61: I don't think this is necessary since we set:
https://github.com/canonical/gateway-api-integrator-operator/blob/52b18b5c0a5f1c5002d9977854b815f6aa491ae8/charmcraft.yaml#L36

Meaning that Juju won't let the charm being deployed if the controller is not >3.3

src/tls_relation.py
```diff
@@ -5,35 +5,54 @@
...
+        certificate: The certificate in PEM format.
+
+    Returns:
+        The hostname the certificate is issue to.
+    """
+    decoded_cert = x509.load_pem_x509_certificate(certificate.encode())
+
+    common_name_attribute = decoded_cert.subject.get_attributes_for_oid(NameOID.COMMON_NAME)
+    if not common_name_attribute:
+        return ""
```
gregory-schiano, line 38: Not sure we should return an empty string.
If you think we should, then when you call that method you should be defensive

src/tls_relation.py
```diff
@@ -5,35 +5,54 @@
...
         """Init method for the class.
 
         Args:
-            model: The charm model used to get the relations and secrets.
+            certificates: The TLS certificates requirer library.
         """
-        self.charm_model = model
-        self.charm_app = model.app
+        self.certificates = certificates
+        self.model = self.certificates.model
```
gregory-schiano, line 53: It's a weird way to access the model, very indirect. Why removing the model from the constructor ?

src/tls_relation.py
```diff
@@ -44,161 +63,87 @@ def generate_password(self) -> str:
...
 
-    def certificate_relation_created(self, hostname: str, tls_integration: Relation) -> None:
+    def certificate_relation_created(self, hostname: str) -> None:
         """Handle the TLS Certificate created event.
 
         Args:
             hostname: Certificate's hostname.
-            tls_integration: The tls certificates integration.
         """
+        # At this point, TLSInformation state component should already be initialized
```
gregory-schiano, line 87: Perhaps adding an assert before casting ?

src/tls_relation.py
```diff
@@ -207,9 +152,20 @@ def _get_private_key(self, hostname: str) -> Dict[str, str]:
...
-        private_key_dict = {}
-        if JujuVersion.from_environ().has_secrets:
-            secret = self.charm_model.get_secret(label=f"private-key-{hostname}")
-            private_key_dict["key"] = secret.get_content()["key"]
-            private_key_dict["password"] = secret.get_content()["password"]
-        return private_key_dict
+        secret = self.model.get_secret(label=f"private-key-{hostname}")
+        private_key = secret.get_content()["key"]
+        password = secret.get_content()["password"]
+        return (private_key, password)
```
gregory-schiano, line 158: I think you could create a `KeyPair` Namedtuple as this is often used at many places sometimes as a dict sometimes as a tuple

Thanhphan1147: Makes sense, I updated the method to return a `KeyPair` named tuple.

src/tls_relation.py
Thanhphan1147: I added an assert here.

src/tls_relation.py
Thanhphan1147: I was aiming to reduce the number of dependencies for the initialization of this module. I added back the `model` parameter.

src/tls_relation.py
Thanhphan1147: I think the idea behind this was that we can trust that the provider will send us valid certificates so we don't need to be defensive. I've thought about it for a while and I agree that doing it like this is a bit fragile since now the method is used in way more places than in the `nginx-ingress-integrator` charm and the output of this is used to lookup secrets and re-requesting new certs as well so raising an exception here makes sense for me. I've add that code in and handle the thrown exception in `validation.py`

src/charm.py
Thanhphan1147: we are replying on the `from_charm` method to check for the existence of the `certificates` integration and raises an exception which is handled by the `@validate_config_and_integration` decorator. I've added another class method called `validate` and used it here to be more explicit on what we're trying to do.

src/charm.py
Thanhphan1147: we are replying on the `from_charm` method to check for the existence of the `certificates` integration and raises an exception which is handled by the `@validate_config_and_integration` decorator. I've added another class method called `validate` and used it here to be more explicit on what we're trying to do.

src/charm.py
Thanhphan1147: we are replying on the `from_charm` method to check for the existence of the `certificates` integration and raises an exception which is handled by the `@validate_config_and_integration` decorator. I've added another class method called `validate` and used it here to be more explicit on what we're trying to do.

src/charm.py
Thanhphan1147: Same comments above

src/charm.py
Thanhphan1147: Same comments above

src/charm.py
Thanhphan1147: Same comments above

src/resource_manager/gateway.py
Thanhphan1147: I rephrased this to be more explicit.

tests/unit/conftest.py
Thanhphan1147: I rephrased this to be more explicit.

src/tls_relation.py
```diff
@@ -5,200 +5,169 @@
...
-            model: The charm model used to get the relations and secrets.
-        """
-        self.charm_model = model
-        self.charm_app = model.app
+class KeyPair(typing.NamedTuple):
+    """Stores a private key and encryption password.
 
-    def generate_password(self) -> str:
-        """Generate a random 12 character password.
+    Attrs:
```
yanksyoon, line 32: I think `Attrs` doesn't work with docstrings-complete. `Attributes` should work though :D

yanksyoon (dismissed):
LGTM! Minor nits :D

src/state/tls.py
```diff
@@ -22,55 +25,61 @@ class TLSInformation:
...
         tls_certs = {}
         tls_keys = {}
         secret_resource_name_prefix = f"{charm.app.name}-secret"
 
-        for key, value in tls_requirer_integration.data[charm.app].items():
-            if key.startswith("chain-"):
-                hostname = key.split("-", maxsplit=1)[1]
-                tls_certs[hostname] = value
-
-                if JujuVersion.from_environ().has_secrets:
```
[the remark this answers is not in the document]
gregory-schiano: That's already set to juju > 3.3

src/tls_relation.py
cbartz: src-docs doesn't render it properly if its named `Attrs`

src/tls_relation.py
```diff
@@ -85,7 +107,9 @@ def certificate_relation_created(self, hostname: str) -> None:
             hostname: Certificate's hostname.
         """
         # At this point, TLSInformation state component should already be initialized
-        tls_integration = typing.cast(Relation, self.model.get_relation(self.integration_name))
+        tls_integration = self.model.get_relation(self.integration_name)
+        assert tls_integration
```
cbartz, line 111: this will probably lead to a bandit warning once linting is enabled: https://bandit.readthedocs.io/en/latest/plugins/b101_assert_used.html

cbartz (dismissed):

src/tls_relation.py
gregory-schiano: We can use `raise AssertionError` instead

src/state/tls.py
```diff
@@ -22,55 +25,61 @@ class TLSInformation:
...
 
         return cls(
             secret_resource_name_prefix=secret_resource_name_prefix,
-            tls_requirer_integration=tls_requirer_integration,
             tls_certs=tls_certs,
             tls_keys=tls_keys,
         )
+
+    @classmethod
+    def validate(cls, charm: ops.CharmBase) -> None:
```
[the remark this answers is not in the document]
gregory-schiano: As discussed, I'm much more in favor of validation error to be exceptions than returning `None` that will require the caller to take a defensive approach

src/resource_manager/resource_manager.py
Thanhphan1147: I rephrased this to be more explicit.

src/tls_relation.py
Thanhphan1147: I renamed all `Attrs` to `Attributes`

src/tls_relation.py
Thanhphan1147: Thanks, I updated the code to raise AssertionError.

cbartz (dismissed):

cbartz (approved):

gregory-schiano (approved):
LGTM

PR #56, opened 2024-07-27
ISD-2020 Request cert when config changed

Thanhphan1147:
Currently the charm does not re-request a new certificate when external-hostname config is updated. This PR adds this functionality. When the `external-hostname` config is changed, the config value is checked against the hostname of all the listeners in the current gateway resource. Currently if:

- The listeners have different hostnames or invalid hostnames
- The configured hostname is different from the hostname of the listeners

The charm will revoke all provider certificates and request a new one with the appropriate hostname.

### Checklist


src/resource_manager/gateway.py
```diff
@@ -213,3 +213,16 @@ def gateway_address(self, name: str) -> typing.Optional[str]:
...
+    def current_gateway_resource(self) -> typing.Optional[GenericNamespacedResource]:
+        """Return the current gateway resource.
+
+        Returns:
+            typing.Optional[GenericNamespacedResource]: The gateway resource
+            or None if the resource does not exists or number of gateway resources != 1.
+        """
+        gateway_resources = self._list_resource()
+        logger.info("%r", gateway_resources)
+        if not gateway_resources or len(gateway_resources) != 1:
```
weiiwang01, line 226: Is it possible to have more than one gateway resource, since there can be only one integration for this charm?

Thanhphan1147: Yes it's possible to have more than one gateway resource with the same label selector since we don't have control over the k8s namespace so anyone can create/modify the gateway resource that's managed by the charm. My thought is that in cases like these we can't do much until the gateway resource is updated to its expected state.

weiiwang01 (approved):

src/charm.py
```diff
@@ -380,6 +394,31 @@ def _set_status_gateway_address(
...
+            labels=self._labels,
+            client=client,
+        )
+        gateway = gateway_resource_manager.current_gateway_resource()
+        if not gateway:
+            return False
+
+        gateway_listeners = gateway.spec["listeners"]
+        listener_hostnames = [listener["hostname"] for listener in gateway_listeners]
+        if len(set(listener_hostnames)) == 1 and config.external_hostname == listener_hostnames[0]:
```
amandahla, line 417: Can someone change the gateway-class and this returns True, revoking certificates even if is not needed?

Thanhphan1147: The method only returns true if the hostname in the listeners is different from what is configured, so i think it doesn't matter which config triggers the hook method if revocation is not needed then the charm should do nothing.

amandahla (approved):

src/charm.py
amandahla: Lets keep an eye on that to guarantee that a unrelated config-change triggers a certificate revocation.

Thanhphan1147: Yep, I'll keep that in mind.

PR #256, opened 2026-06-01
feat: Add gateway-route v1 interface

swetha1654:
#### What this PR does

- Added gateway-route v1 interface.

#### Why we need it

Implements [spec](https://docs.google.com/document/d/1gzme7jCqZ0E8jj9rSsF8oiVaOrttpnkbPT1bh69t3es/edit?tab=t.0).

#### Checklist

### DIFF
--- a/gateway-api-integrator/lib/charms/gateway_api_integrator/v1/gateway_route.py
+++ b/gateway-api-integrator/lib/charms/gateway_api_integrator/v1/gateway_route.py
@@ -1,0 +1,412 @@
+# Copyright 2026 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""gateway-route interface library v1.
+
+This library implements the gateway-route interface.
+
+The requirer publishes hostname information. The provider publishes
+gateway resource details (gateway name, model/namespace, HTTPS mode)
+so the requirer can reference the correct Gateway resource.
+
+With the v1 interface, the requirer is expected to create and manage
+its own HTTP/TCP/UDP route resources referencing the provider's Gateway.
+
+## Getting Started
+
+Fetch the library:
+
+```shell
+charmcraft fetch-lib charms.gateway_api_integrator.v1.gateway_route
+```
+
+### Requirer usage
+
+In the `charmcraft.yaml` of the charm, add the following:
+
+```yaml
+requires:
+    gateway-route:
+        interface: gateway-route
+```
+
+```python
+from charms.gateway_api_integrator.v1.gateway_route import (
+    GATEWAY_ROUTE_RELATION_NAME,
+    GatewayRouteRequirer,
+    GatewayRouteProviderAppData,
+)
+
+class IngressConfiguratorCharm(CharmBase):
+    def __init__(self, *args):
+        super().__init__(*args)
+        self.gateway_route = GatewayRouteRequirer(self)
+        self.framework.observe(
+            self.on[GATEWAY_ROUTE_RELATION_NAME].relation_changed, self._on_gateway_route_changed
+        )
+        self.framework.observe(
+            self.on[GATEWAY_ROUTE_RELATION_NAME].relation_broken, self._on_gateway_route_changed
+        )
+
+    def _on_gateway_route_changed(self, event):
+        provider_data = self.gateway_route.get_provider_data()
+        if provider_data:
+            # Use provider_data.gateway_name, provider_data.model_name, provider_data.https_mode
+            ...
+```
+
+### Provider usage
+
+In the `charmcraft.yaml` of the charm, add the following:
+
+```yaml
+provides:
+    gateway-route:
+        interface: gateway-route
+```
+
+```python
+from charms.gateway_api_integrator.v1.gateway_route import (
+    GATEWAY_ROUTE_RELATION_NAME,
+    GatewayRouteProvider,
+    HttpsMode,
+    GatewayRouteRequirerAppData,
+)
+
+class GatewayAPIIntegratorCharm(CharmBase):
+    def __init__(self, *args):
+        super().__init__(*args)
+        self.gateway_route = GatewayRouteProvider(self)
+        self.framework.observe(
+            self.on[GATEWAY_ROUTE_RELATION_NAME].relation_changed, self._on_gateway_route_changed
+        )
+        self.framework.observe(
+            self.on[GATEWAY_ROUTE_RELATION_NAME].relation_broken, self._on_gateway_route_changed
+        )
+
+    def _on_gateway_route_changed(self, event):
+        for relation in self.model.relations[GATEWAY_ROUTE_RELATION_NAME]:
+            requirer_data = self.gateway_route.get_requirer_data(relation)
+            if requirer_data:
+                # Use requirer_data.hostname, requirer_data.additional_hostnames
+                ...
+        # Publish provider data to all relations
+        self.gateway_route.publish_provider_data(
+            gateway_name=self.app.name,
+            model_name=self.model.name,
+            https_mode=HttpsMode.ENFORCED,
+        )
+```
+"""
+
+import logging
+from enum import StrEnum
+from typing import Annotated
+
+from ops import CharmBase
+from ops.framework import Object
+from ops.model import (
+    Relation,
+    RelationDataTypeError,
+)
+from pydantic import BeforeValidator, Field, ValidationError
+from pydantic.dataclasses import dataclass
+from validators import domain
+
+# The unique Charmhub library identifier, never change it
+LIBID = "53fdf90019a7406695064ed1e3d2708f"
+
+# Increment this major API version when introducing breaking changes
+LIBAPI = 1
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 0
+
+logger = logging.getLogger(__name__)
+GATEWAY_ROUTE_RELATION_NAME = "gateway-route"
+
+
+class GatewayRouteInvalidRelationDataError(Exception):
+    """Raised when relation data validation for gateway-route fails."""
+
+
+def valid_fqdn(value: str) -> str:
+    """Validate if value is a valid FQDN (TLDs alone are not allowed).
+
+    Args:
+        value: The value to validate.
+
+    Raises:
+        ValueError: When value is not a valid domain.
+    """
+    if not bool(domain(value)):
+        raise ValueError(f"Invalid domain: {value}")
+    return value
+
+
+# --- Data models ---
+
+
+class HttpsMode(StrEnum):
+    """HTTPS mode for the gateway.
+
+    Attrs:
+        DISABLED: TLS is not configured; only HTTP listener exists.
+        ENABLED: TLS is configured; both HTTP and HTTPS listeners exist.
+        ENFORCED: TLS is configured and enforce-https is true; HTTP redirects to HTTPS.
+    """
+
+    DISABLED = "disabled"
+    ENABLED = "enabled"
+    ENFORCED = "enforced"
+
+
+@dataclass
+class GatewayRouteRequirerAppData:
+    """Requirer application databag schema.
+
+    The requirer provides hostname information so the provider can
+    configure TLS certificates and DNS records.
+
+    Attributes:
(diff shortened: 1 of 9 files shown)


gateway-api-integrator/lib/charms/gateway_api_integrator/v1/gateway_route.py
```diff
@@ -0,0 +1,412 @@
...
+        try:
+            app_data = GatewayRouteProviderAppData(
+                gateway_name=gateway_name,
+                model_name=model_name,
+                https_mode=https_mode,
+            )
+            relation.save(app_data, self.charm.app)
+        except (ValidationError, RelationDataTypeError) as exc:
+            logger.error("Failed to publish provider data to relation %s.", relation.id)
+            raise GatewayRouteInvalidRelationDataError(
```
javierdelapuente, line 295: As the charm is still not there I cannot see what happens, but if the third relation out of five fails, will it work? what would happen to the two last relations?

javierdelapuente (approved):

gateway-api-integrator/lib/charms/gateway_api_integrator/v1/gateway_route.py
```diff
@@ -0,0 +1,412 @@
...
+        self._valid_relations = []
+        for relation in self.relations:
+            try:
+                data = self.get_requirer_data(relation)
+                results[relation.id] = data
+                self._valid_relations.append(relation)
+            except GatewayRouteInvalidRelationDataError:
+                logger.error(
+                    "Skipping relation %s: invalid data", relation.id
+                )
```
seb4stien, line 252: Maybe we should reduce the try block to the minimum.

```suggestion
            try:
                data = self.get_requirer_data(relation)
            except GatewayRouteInvalidRelationDataError:
                logger.error(
                    "Skipping relation %s: invalid data", relation.id
                )
            else:
                results[relation.id] = data
                self._valid_relations.append(relation)
```

gateway-api-integrator/lib/charms/gateway_api_integrator/v1/gateway_route.py
```diff
@@ -0,0 +1,412 @@
...
+        """
+        try:
+            app_data = GatewayRouteProviderAppData(
+                gateway_name=gateway_name,
+                model_name=model_name,
+                https_mode=https_mode,
+            )
+            relation.save(app_data, self.charm.app)
+        except (ValidationError, RelationDataTypeError) as exc:
+            logger.error("Failed to publish provider data to relation %s.", relation.id)
```
seb4stien: Isn't it redundant with the exception below?

gateway-api-integrator/lib/charms/gateway_api_integrator/v1/gateway_route.py
```diff
@@ -0,0 +1,412 @@
...
+            return
+
+        try:
+            app_data = GatewayRouteRequirerAppData(
+                hostname=hostname,
+                additional_hostnames=additional_hostnames or [],
+            )
+            relation.save(app_data, self.charm.app)
+        except (ValidationError, RelationDataTypeError) as exc:
+            logger.error("Failed to publish requirer data.")
```
seb4stien: Isn't it redundant with raising the exception below?

seb4stien (approved):

gateway-api-integrator/lib/charms/gateway_api_integrator/v1/gateway_route.py
swetha1654: i will use `continue` instead of else since its slightly more readable

f-atwi (approved):

gateway-api-integrator/lib/charms/gateway_api_integrator/v1/gateway_route.py
swetha1654: good point, that was not handled. I updated the code to process the remaining 2 relations too. 

on the other hand, if validationError / RelationDataTypeError occur, its mostly because the provider is pushing something wrong and it will fail on all relations anyway since we are pushing the same data everywhere, so technically its not really required to process all relations. But i still think its better to process all relations anyway.

gateway-api-integrator/lib/charms/gateway_api_integrator/v1/gateway_route.py
```diff
@@ -0,0 +1,423 @@
...
+
+        Raises:
+            GatewayRouteInvalidRelationDataError: When data validation fails.
+        """
+        try:
+            return relation.load(GatewayRouteRequirerAppData, relation.app)
+        except ValidationError as exc:
+            raise GatewayRouteInvalidRelationDataError(
+                f"gateway-route data validation failed for relation {relation.id}"
+            ) from exc
```
Thanhphan1147: Do we even need this now since it's just a wrapper around relation.load()?

gateway-api-integrator/lib/charms/gateway_api_integrator/v1/gateway_route.py
```diff
@@ -0,0 +1,423 @@
...
+        results: dict[int, GatewayRouteRequirerAppData] = {}
+        self._valid_relations = []
+        for relation in self.relations:
+            try:
+                data = self.get_requirer_data(relation)
+            except GatewayRouteInvalidRelationDataError:
+                logger.error(
+                    "Skipping relation %s: invalid data", relation.id
+                )
+                continue
```
Thanhphan1147: We probably need a way to track invalid relation IDs as well, and reflect the status that the relation data is invalid

gateway-api-integrator/lib/charms/gateway_api_integrator/v1/gateway_route.py
```diff
@@ -0,0 +1,423 @@
...
+            except GatewayRouteInvalidRelationDataError:
+                logger.error(
+                    "Failed to publish provider data to relation %s", relation.id
+                )
+                failed_relations.append(relation.id)
+
+        if failed_relations:
+            raise GatewayRouteInvalidRelationDataError(
+                f"Failed to publish provider data to relations: {failed_relations}"
+            )
```
Thanhphan1147, line 297: Are we catching and raising the same exception? (GatewayRouteInvalidRelationDataError)

Thanhphan1147: We also should not block/error the provider if publishing to this databag fail I think

gateway-api-integrator/lib/charms/gateway_api_integrator/v1/gateway_route.py
```diff
@@ -0,0 +1,423 @@
...
+            app_data = GatewayRouteProviderAppData(
+                gateway_name=gateway_name,
+                model_name=model_name,
+                https_mode=https_mode,
+            )
+            relation.save(app_data, self.charm.app)
+        except (ValidationError, RelationDataTypeError) as exc:
+            raise GatewayRouteInvalidRelationDataError(
+                "Failed to publish provider relation data."
+            ) from exc
```
Thanhphan1147: likewise, do we need this as this is just a wrapper for relation.save? can we maybe simplify it and move it to `publish_provider_data`?

gateway-api-integrator/lib/charms/gateway_api_integrator/v1/gateway_route.py
```diff
@@ -0,0 +1,423 @@
...
+            hostname: The hostname for the service.
+            additional_hostnames: Additional hostnames for the service.
+
+        Raises:
+            GatewayRouteInvalidRelationDataError: When data validation fails.
+        """
+        relation = self.relation
+        if not relation or not self.charm.unit.is_leader():
+            return
+
```
Thanhphan1147: (not blocking)
we can maybe simplify with 
```python
if (relation := self.relation) and (not self.charm.unit.is_leader()):
  ...
````

File: gateway-api-integrator/lib/charms/gateway_api_integrator/v1/gateway_route.py [hunk 10]
```diff
@@ -0,0 +1,423 @@
...
+        Returns:
+            Validated provider data or None if not yet available.
+
+        Raises:
+            GatewayRouteInvalidRelationDataError: When data validation fails.
+        """
+        relation = self.relation
+        if not relation or not relation.app:
+            return None
```
Thanhphan1147: same comments as above

Thanhphan1147 (commented):
Some initial reviews

gateway-api-integrator/lib/charms/gateway_api_integrator/v1/gateway_route.py
```diff
@@ -0,0 +1,423 @@
...
+    can construct HTTPRoute resources referencing the correct Gateway.
+
+    Attributes:
+        gateway_name: The name of the Gateway resource managed by the provider.
+        model_name: The Juju model (K8s namespace) where the Gateway resource lives.
+        https_mode: The HTTPS mode indicating how the provider handles TLS.
+    """
+
+    gateway_name: str = Field(description="The name of the Gateway resource.")
+    model_name: str = Field(
```
f-atwi: do you think `gateway_model` is a more fitting name?

swetha1654: yes it would, updated

gateway-api-integrator/lib/charms/gateway_api_integrator/v1/gateway_route.py
swetha1654: yeah we don't, makes the code simpler. Removing

gateway-api-integrator/lib/charms/gateway_api_integrator/v1/gateway_route.py
swetha1654: agreed and moved

gateway-api-integrator/lib/charms/gateway_api_integrator/v1/gateway_route.py
swetha1654: @Thanhphan1147 why would we need to track the invalid relations? Currently the provider logs any relation with invalid data and then ignores it. Do you think the API should change to return a list of the invalid relations too?

gateway-api-integrator/lib/charms/gateway_api_integrator/v1/gateway_route.py
swetha1654: @Thanhphan1147 no we are currently iterating through all the relations on the provider side, storing failed relations in a list, and in the end if the list isn't empty, we are raising `GatewayRouteInvalidRelationDataError` with all the relations. 

This is done to avoid failing to process all the remaining relations if one of them in the middle fails.

gateway-api-integrator/lib/charms/gateway_api_integrator/v1/gateway_route.py
Thanhphan1147: I don't remember what we discussed, but shouldn't the provider not send over the details to bind to the listeners if the requirer has sent invalid data? Because if not then we might as well send the gateway and namespace information in the provider's `__init__` method I think.

swetha1654: @Thanhphan1147 the provider only iterates through the `_valid_relations` which is computed at `get_requirer_data`. 
Note: This code assumes that the `get_requirer_data` always runs before `publish_provider_data`, and i think this coupling is fine for holistic pattern charms. 

I've also documented the same in the docstring

gateway-api-integrator/lib/charms/gateway_api_integrator/v1/gateway_route.py
Thanhphan1147: This is related to my other reply I think. I think we can either:

1. Send the gateway and model information in a "best-effort" way where we do it in `__init__`
2. Keep track of the invalid relation IDs and only provide the gateway and model information for the relations with valid data.

In either cases I think we don't need to raise any exceptions if `relation.save` fails since I guess it'll be a permission issues most of the time and we should not block the entire charm due to one relation.

Thanhphan1147 (approved):

PR #260, opened 2026-06-10
feat: Add support for gateway-route v1

swetha1654:
#### Checklist


gateway-api-integrator/src/resource_manager/gateway.py
```diff
@@ -225,7 +230,7 @@ def gateway_address(self, name: str) -> typing.Optional[str]:
                     for addr in gateway.status["addresses"]  # type: ignore
                 ]
                 if gateway_addresses:
-                    gateway_address = ",".join(gateway_addresses)
+                    gateway_address = gateway_addresses[0]
```
seb4stien: We don't know which IP would be returned. I suggest to still return a list and block on the requirer side saying multiple IPs are not supported yet.

seb4stien: In fact, it could make more sense to block on the provider side (as it would also impact the certificates logic).

gateway-api-integrator/src/state/charm_state.py
```diff
@@ -54,20 +56,41 @@ class ProxyMode(StrEnum):
...
+    requires_ip_certificate: bool = Field()
+    hostnames: set[typing.Annotated[str, BeforeValidator(valid_fqdn)]] = Field(default_factory=set)
+
+    @property
+    def hostname(self) -> str | None:
+        """Return external-hostname config for the ingress relation."""
+        if self.proxy_mode != ProxyMode.INGRESS:
+            raise ValueError("hostname property can only be accessed when proxy_mode is INGRESS")
+        if len(self.hostnames) > 1:
+            raise ValueError("Expected a single hostname, but multiple were found")
```
seb4stien: Aren't these ones RuntimeErrors?

gateway-api-integrator/src/state/config.py
```diff
@@ -54,20 +56,41 @@ class ProxyMode(StrEnum):
     INACTIVE = "inactive"
 
 
+INGRESS_RELATION = "gateway"
+GATEWAY_ROUTE_RELATION = "gateway-route"
+
+
 @dataclass(frozen=True)
 class CharmConfig:
```
seb4stien: Is it a config or a State?

gateway-api-integrator/src/resource_manager/gateway.py
Thanhphan1147: iiuc this method is only used to set the status of the charm so I think the operator should know about all of the addresses to connect to the gateway ( if there are many )

gateway-api-integrator/src/state/charm_state.py
```diff
@@ -54,20 +56,41 @@ class ProxyMode(StrEnum):
...
     enforce_https: bool = Field()
     proxy_mode: ProxyMode = Field()
+    requires_ip_certificate: bool = Field()
+    hostnames: set[typing.Annotated[str, BeforeValidator(valid_fqdn)]] = Field(default_factory=set)
+
+    @property
+    def hostname(self) -> str | None:
+        """Return external-hostname config for the ingress relation."""
+        if self.proxy_mode != ProxyMode.INGRESS:
+            raise ValueError("hostname property can only be accessed when proxy_mode is INGRESS")
```
Thanhphan1147: I think it's an anti-pattern for properties to raise exceptions

gateway-api-integrator/src/state/charm_state.py
```diff
@@ -95,15 +118,26 @@ def from_charm_and_providers(
...
 
         proxy_mode = cls._validate_state(
-            ingress_provider.relations, gateway_route_provider.relation
+            ingress_provider.relations, gateway_route_provider.relations
         )
         gateway_class_name = typing.cast(str, charm.config.get("gateway-class"))
+
+        # external-hostname is an ingress-mode-only config option. In gateway-route
+        # mode hostnames come from the relation data, so setting it is a misconfiguration.
+        config_external_hostname = typing.cast(str | None, charm.config.get("external-hostname"))
```
Thanhphan1147: Since we defined a clear separation for ingress and gateway-route, it might be better to split the charm state into 2 components and try to isolate the two modes as much as possible to avoid complex logic

gateway-api-integrator/src/state/charm_state.py
```diff
@@ -115,35 +149,80 @@ def from_charm_and_providers(
...
     ) -> ProxyMode:
         """Validate the charm config state.
 
         Raises:
             InvalidCharmConfigError: When the charm config is invalid.
         """
         is_ingress_related = bool(ingress_relations)
-        is_gateway_route_related = gateway_route_relation is not None
+        is_gateway_route_related = bool(gateway_route_relations)
         if is_ingress_related and is_gateway_route_related:
```
Thanhphan1147, line 230: similar to my comment above, I think we can save ourselves quite a bit by isolating the 2 modes

gateway-api-integrator/src/state/tls.py
```diff
@@ -42,55 +39,72 @@ class TLSInformation:
...
+    def hostnames(self) -> list[str]:
+        """Get the list of hostnames for the TLS information."""
+        return list(self.tls_certs.keys())
 
     @classmethod
     def from_charm(
         cls,
         charm: ops.CharmBase,
-        hostname: str | None,
+        config: CharmConfig,
```
Thanhphan1147: Can we also pass the list of hostnames instead of the entire charmconfig to minimize the dependencies between state components? ( since it will impact ordering, and couple them together when you initialize them )

gateway-api-integrator/src/resource_manager/gateway.py
Thanhphan1147: Just realized later in the review that this is also used to get the certificates, still I think we need to request a certificate for each LB address since we don't know which one the operator will use to connect

gateway-api-integrator/src/state/tls.py
```diff
@@ -42,55 +39,72 @@ class TLSInformation:
...
                 if chain[0] != cert.certificate:
                     chain.reverse()
-                tls_certs[hostname] = "\n\n".join([str(cert) for cert in chain])
-                tls_keys[hostname] = str(certificates.private_key)
+                tls_certs[cn] = "\n\n".join([str(c) for c in chain])
+                tls_keys[cn] = str(certificates.private_key)
+
+        # In the IP SAN case the certificate is requested only after the gateway IP
+        # is known, so it may not be issued yet. Signal "not ready" instead of
+        # raising, allowing the charm to converge once the certificate is available.
```
Thanhphan1147: We can still raise an exception here and handle it in the charm code to set the status to Waiting to be aligned with how it's done in other places.

gateway-api-integrator/src/charm.py
```diff
@@ -231,18 +217,42 @@ def _on_dns_record_relation_joined(self, _: RelationJoinedEvent) -> None:
...
+        self.certificates.certificate_requests = csrs
+        self.certificates.sync()
+
+    def _determine_https_mode(self, config: CharmConfig, has_tls_relation: bool) -> HttpsMode:
+        """Determine the HTTPS mode based on config and TLS relation presence."""
+        if config.enforce_https:
+            return HttpsMode.ENFORCED
+        if has_tls_relation:
+            return HttpsMode.ENABLED
+        return HttpsMode.DISABLED
```
Thanhphan1147: nit: Maybe
```suggestion
        return HttpsMode.ENABLED if has_tls_relation else HttpsMode.DISABLED
```

gateway-api-integrator/src/charm.py
```diff
@@ -259,74 +269,155 @@ def _reconcile(self) -> None:
...
         gateway_resource_manager = GatewayResourceManager(
             labels=self._labels,
             client=client,
         )
+
+        # NOTE: When no hostname is configured we serve the gateway with an IP SAN
+        # certificate. The gateway IP is only known once the gateway exists and the
+        # load balancer has allocated an address. So the first pass creates the
+        # gateway resource without the IP SAN certificate. Subsequent hooks will
+        # add the IP SAN certificate.
```
Thanhphan1147: I think we should set the charm to waiting while the certificate is not there ( I think we're already setting the charm to waiting if the gateway address is not available )

gateway-api-integrator/src/charm.py
```diff
@@ -259,74 +269,155 @@ def _reconcile(self) -> None:
...
+        gateway_resource_information: GatewayResourceInformation,
+        gateway_resource_manager: GatewayResourceManager,
+    ) -> None:
+        """Publish provider data for gateway-route relations when gateway address is ready."""
+        # Refresh the set of valid relations to publish to.
+        self._gateway_route_provider.get_requirer_data()
+
+        # Clean up unused resources
+        HTTPRouteResourceManager(self._labels, client).cleanup_resources(exclude=[])
+        ServiceResourceManager(self._labels, client).cleanup_resources(exclude=[])
```
Thanhphan1147: I might be wrong but I'd imagine this is for when switching from ingress to gateway-route, though I think I prefer not to let ingress-specific notions to creep into the logic of gateway-route and keep them separated ( it doesn't make sense for gateway-route to have to clean up resources created by ingress - it should be ingress that has to do the cleanup itself I think )

Thanhphan1147 (commented):
First pass without looking at the tests, I left some comments

gateway-api-integrator/src/resource_manager/gateway.py
swetha1654: @Thanhphan1147 I spoke to @seb4stien about it, its not just requesting a certificate but also creating the ingress URL in case of ingress mode and providing the address back in case of gateway route. I don't expect us to anything other than a single ipv4 address in most cases, so for now we decided to block if we get anything else saying this is not supported, for simplicity.

gateway-api-integrator/src/state/charm_state.py
swetha1654: removed the property

gateway-api-integrator/src/state/charm_state.py
swetha1654: There are only 4 properties and only one of the property (hostname for ingress / hostnames for gateway-route) differs for ingress and gateway-route. So I have proposed the common charm_state which then splits into ingress and gatewayroute state classes.

gateway-api-integrator/src/charm.py
swetha1654: @Thanhphan1147 its not just for switching between ingress and gateway-route, but also cleaning up resources from an older deployment. I don't think juju refresh clears up and k8s resources managed by gateway API.

swetha1654: > it doesn't make sense for gateway-route to have to clean up resources created by ingress - it should be ingress that has to do the cleanup itself

If we switch from ingress to gateway-route mode, we will never hit that logic

swetha1654: moved the cleanup logic out of gateway-route mode

gateway-api-integrator/lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -0,0 +1,1824 @@
+# Copyright 2021 Canonical Ltd.
```
seb4stien: IIRC it's a test dependency, I wonder if it will be packed in the charm and if we are ok with that as it's unrelated to the charm itself.
In fact, there's a corresponding charmlib, so you may be able to properly define it as a dev requirement and it wouldn't show up in the repo.

gateway-api-integrator/src/resource_manager/gateway.py
```diff
@@ -224,8 +230,32 @@ def gateway_address(self, name: str) -> typing.Optional[str]:
                     addr["value"]  # type: ignore
                     for addr in gateway.status["addresses"]  # type: ignore
                 ]
-                if gateway_addresses:
-                    gateway_address = ",".join(gateway_addresses)
+
+                ipv4_addresses = []
+                for address in gateway_addresses:
+                    try:
+                        parsed_address = ipaddress.ip_address(address)
```
seb4stien, line 237: ip_adress will return successfully if provided a valid IPv6 address, no you may be missing a check her if you want to enforce ipv4.

seb4stien: My bad. I did not see the test below :)

gateway-api-integrator/src/state/charm_state.py
```diff
@@ -54,20 +56,45 @@ class ProxyMode(StrEnum):
...
     gateway_class_name: str = Field(min_length=1)
     enforce_https: bool = Field()
     proxy_mode: ProxyMode = Field()
+    requires_ip_certificate: bool = Field()
+    hostnames: set[
+        typing.Annotated[
+            str,
+            BeforeValidator(valid_fqdn),
+        ]
+    ] = Field(default_factory=set)
```
seb4stien, line 87: While the "set" will provide uniqueness, it ma lose the ordering, don't we want to preserve the ordering to have the "hostname" first and the additional hostnames after?

gateway-api-integrator/src/state/charm_state.py
```diff
@@ -54,20 +56,45 @@ class ProxyMode(StrEnum):
...
+
+    @field_validator("hostnames", mode="after")
+    @classmethod
+    def validate_ingress_hostnames_count(
+        cls, hostnames: set[str], info: ValidationInfo
+    ) -> set[str]:
+        """Ingress mode supports at most one hostname."""
+        if info.data.get("proxy_mode") == ProxyMode.INGRESS and len(hostnames) > 1:
+            raise ValueError("ingress mode supports at most one hostname")
```
seb4stien: Don't refer to "modes" in user facing messages. Maybe something like "When related with 'ingress', the charm support at most"...

gateway-api-integrator/src/state/charm_state.py
```diff
@@ -92,18 +119,29 @@ def from_charm_and_providers(
             IngressGatewayRouteConflictError: When both ingress and gateway-route is present.
 
         Returns:
-            CharmConfig: Instance of the charm config state component.
+            CharmState for inactive mode, or a mode-specific subclass for active integrations.
         """
         enforce_https = typing.cast(bool, charm.config.get("enforce-https", True))
-        if charm.model.get_relation(TLS_CERTIFICATES_INTEGRATION) is None and enforce_https:
+        has_tls = charm.model.get_relation(TLS_CERTIFICATES_INTEGRATION) is not None
+        if not has_tls and enforce_https:
```
seb4stien: Maybe me, but I find the logic easier to read that way:
```suggestion
        if enforce_https and not has_tls:
```

gateway-api-integrator/src/state/charm_state.py
```diff
@@ -115,35 +153,81 @@ def from_charm_and_providers(
...
             )
 
-        try:
-            gateway_route_requirer_data = gateway_route_provider.get_data()
-            hostname = gateway_route_requirer_data.application_data.hostname
-        except (GatewayRouteInvalidRelationDataError, GatewayRouteRelationMissingError):
-            hostname = typing.cast(str | None, charm.config.get("external-hostname"))
+        if proxy_mode == ProxyMode.INGRESS and has_tls and not config_external_hostname:
+            raise HostnameMissingError(
+                "external-hostname must be set in ingress mode "
```
seb4stien: Please remove the reference to "mode" (= align with the previous InvalidCharmConfigError message).

gateway-api-integrator/src/state/charm_state.py
```diff
@@ -115,35 +153,81 @@ def from_charm_and_providers(
...
+        return None
+
+    @staticmethod
+    def get_gateway_route_hostnames(gateway_route_provider: GatewayRouteProvider) -> set[str]:
+        """Return hostname set used for gateway-route mode certificate requests."""
+        hostnames: set[str] = set()
+        requirer_data = gateway_route_provider.get_requirer_data()
+        for data in requirer_data.values():
+            if data.hostname:
+                hostnames.add(data.hostname)
```
seb4stien, line 200: Shouldn't we run some valid_fqdn validation here?

gateway-api-integrator/src/charm.py
```diff
@@ -134,17 +141,29 @@ def __init__(self, *args) -> None:  # type: ignore[no-untyped-def]
...
+        try:
+            charm_state = CharmState.from_charm_and_providers(
+                self,
+                self.available_gateway_classes(),
+                self._ingress_provider,
+                self._gateway_route_provider,
+            )
+            csrs = [
+                CertificateRequestAttributes(common_name=hostname, sans_dns=[hostname])
+                for hostname in sorted(charm_state.hostnames)
```
seb4stien, line 159: I'm not sure about the logic here. If I understand correctly, we request a single certificate for all hostname accross all relations? Is it what we want? I thought it would be one certificate per relation.

gateway-api-integrator/src/charm.py
```diff
@@ -253,100 +275,190 @@ def _reconcile(self) -> None:
...
+            hostnames: Hostnames to publish as DNS records.
         """
         relation = self.model.get_relation(self.dns_record_requirer.relation_name)
-        if not relation or not external_hostname:
+        if not relation:
             return
+
+        sorted_hostnames = sorted(hostnames)
+        if not sorted_hostnames:
+            return
```
seb4stien: Nitpicking
```suggestion
        if not hostnames:
           return
        sorted_hostnames = sorted(hostnames)
```

gateway-api-integrator/src/charm.py
```diff
@@ -412,36 +506,25 @@ def _define_ingress_resources_and_publish_url(
...
-        if config.proxy_mode == ProxyMode.GATEWAY_ROUTE:
-            http_route_resource_information = HTTPRouteResourceInformation.from_gateway_route(
-                self._gateway_route_provider,
-                config.hostname,
-            )
-
-        if http_route_resource_information is None:
-            logger.error("Can't determine HTTP route resource information from providers.")
-            return
+        hostname = next(iter(charm_state.hostnames), None)
```
seb4stien, line 545: Is the fallback to "None" a real possibility?

tests/e2e/conftest.py
```diff
@@ -3,25 +3,21 @@
...
-        name (str): The name of the deployed application.
-    """
-
-    name: str
-
+GATEWAY_API_INTEGRATOR_APP_NAME = "gateway-api-integrator"
+INGRESS_CONFIGURATOR_APP_NAME = "ingress-configurator"
+ANY_CHARM_INGRESS_REQUIRER_APP_NAME = "any-charm"
+ANY_CHARM_INGRESS_REQUIRER_SRC = "ingress_requirer.py"
+APT_LIB_SRC = "gateway-api-integrator/lib/charms/operator_libs_linux/v0/apt.py"
```
seb4stien: As mentioned earlier: see if you can use charmlibs instead.

seb4stien (commented):
As discussed during the standup: update the publication track please.

gateway-api-integrator/src/state/charm_state.py
swetha1654: that happens when the charm_state is returned

gateway-api-integrator/src/charm.py
swetha1654: no for the sake of simplicity we are going to request one certificate per hostname and additional_hostname instead of per relation, keeping in line with what HAProxy does. If we don't want this, we will have to change the logic on all ingresses. We can discuss it further

gateway-api-integrator/src/state/charm_state.py
swetha1654: we are sorting it wherever ordering matters

swetha1654: and i don't see why hostname should come first before additional_hostnames in any case

gateway-api-integrator/src/resource_manager/gateway.py
```diff
@@ -34,69 +36,73 @@ class GatewayResourceDefinition(ResourceDefinition):
...
-            tls_secret_name = (
-                f"{tls_information.secret_resource_name_prefix}-{tls_information.hostname}"
-            )
-        super().__init__(gateway_resource_information, charm_config)
-        self.tls_secret_name = tls_secret_name
+            for hostname in tls_information.hostnames:
+                tls_secret_names.append(
+                    f"{tls_information.secret_resource_name_prefix}-{hostname}"
+                )
+        super().__init__(gateway_resource_information, charm_state)
```
javierdelapuente: can we put this line as the first one of the function? it is in between functionality related to tls_secret_names

gateway-api-integrator/src/state/validation.py
```diff
@@ -82,6 +80,10 @@ def wrapper(instance: C, *args: typing.Any) -> None:
                 instance.unit.status = ops.BlockedStatus(str(exc))
                 _clean_up_resources_in_blocked_state(instance)
                 return None
+            except InvalidGatewayAddressError as exc:
+                logger.exception("Gateway reported unsupported address: %s", str(exc))
+                instance.unit.status = ops.BlockedStatus(str(exc))
```
seb4stien: I left a similar comment to Fouad: don't we want to have a human readable message in the status and the full exception in the logs?

seb4stien (approved):

gateway-api-integrator/src/charm.py
swetha1654: yes it is, when TLS is not related and enforce_https is disabled we don't require a hostname.

PR #268, opened 2026-06-26
fix: multiple gateway listeners to fix multiple routes with https (ISD-6028)

f-atwi:
#### What this PR does

Emit one HTTPS listener per hostname. Each listener carries a hostname: field, which Cilium translates to an SNI match, making all filter chains distinct. The listener name follows the convention {gateway_name}-https-{hostname_with_dots_as_hyphens} (e.g. my-gw-https-app-example-com).

#### Why we need it

All HTTPS hostnames were attached as multiple certificateRefs on a single hostname-less HTTPS Gateway listener. Cilium renders each cert as a separate Envoy filter chain with the match {transport_protocol: tls} and no SNI discriminator. When two or more relations are present, Envoy sees duplicate filter-chain matchers and NACKs the entire listener — taking down both :80 and :443.
#### Checklist


gateway-api-integrator/src/resource_manager/gateway.py
```diff
@@ -30,6 +30,22 @@
...
+    where dots in the hostname are replaced with hyphens.
+
+    Args:
+        gateway_name: The name of the Gateway K8s resource.
+        hostname: The hostname for this listener.
+
+    Returns:
+        The listener name.
+    """
+    return f"{gateway_name}-https-{hostname.replace('.', '-')}"
```
swetha1654, line 46: keep in mind that for IP certs, there will be numbers in the listener name. Gateway API allows it, just wanted to bring it to your attention

gateway-api-integrator/tests/unit/test_http_route.py
swetha1654, line 1: can we add some unit tests for truncate_k8s_resource?

swetha1654 (approved):
LGTM, thanks!

gateway-api-integrator/src/resource_manager/gateway.py
```diff
@@ -30,6 +30,38 @@
...
```
seb4stien, line 46: Shouldn't you use truncate_k8s_resource_name here too?

f-atwi: No, we are truncating for k8s resources
this is used in  other places too like the gateway api section name which does not have the same constraints (63 char limit)

seb4stien (approved):

PR #271, opened 2026-07-01
Chore: migrate to charm ci (ISD-6045)

f-atwi:
#### Checklist


gateway-api-integrator/pyproject.toml
```diff
@@ -45,6 +45,7 @@ integration = [
   "allure-pytest-collection-report @ git+https://github.com/canonical/data-platform-workflows@v24.0.0#subdirectory=python/pytest_plugins/allure_pytest_collection_report",
   "jubilant==1.10.0",
   "lightkube",
+  "opcli @ git+https://github.com/canonical/charm-ci.git ; python_version >= '3.12'",
```
javierdelapuente: Before merging I would pin it to the latest rev (still alpha, but at least it does not break you if there is any new version).

artifacts.yaml
```diff
@@ -0,0 +1,13 @@
+# Copyright 2026 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+version: 1
+rocks: []
+charms:
```
javierdelapuente, line 6: aren't we missing the configurator charm?

f-atwi: my next PR is to remove it completely and the integration tests already switched to using ingress-configurator

.github/workflows/integration_test.yaml
```diff
@@ -7,42 +7,28 @@ on:
 
 jobs:
   integration-tests:
-    uses: 
-      canonical/operator-workflows/.github/workflows/integration_test.yaml@main
+    permissions:
+      contents: read
+      packages: write
+      actions: read
+    uses: canonical/charm-ci/.github/workflows/integration-test.yml@main
```
seb4stien: I think we have to pin the version here. Check for the other references to @main too.

seb4stien (approved):

artifacts.yaml
seb4stien: No need to introduce it as we will remove it in next PR indeed.

PR #272, opened 2026-07-01
fix: set SNI in requests in e2e tests to ensure matching hostname with multiple listeners

f-atwi:
#### What this PR does

Switch to using httpx for requiests which allows setting an SNI.
When a hostname is provided, assert_gateway_route_response builds the URL from the hostname (ensuring SNI is set correctly) and mounts the adapter to route the TCP connection to the gateway IP.

#### Why we need it
The multiple-listeners change (rev 163) switched the Gateway resource from a single hostname-less HTTPS listener to per-hostname HTTPS listeners, each with an explicit [hostname](vscode-file://vscode-app/snap/code/247/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) field. Cilium enforces SNI matching against the listener hostname during the TLS handshake.

#### Checklist

### DIFF
--- a/tests/e2e/helper.py
+++ b/tests/e2e/helper.py
@@ -4,15 +4,83 @@
 """Shared helpers for e2e tests."""
 
 import ipaddress
+from urllib.parse import urlparse
 
 import jubilant
 import requests
 import urllib3
+from requests.adapters import DEFAULT_POOLBLOCK, DEFAULT_POOLSIZE, DEFAULT_RETRIES, HTTPAdapter
 from tenacity import retry, retry_if_exception_type, stop_after_delay, wait_fixed
 from urllib3.exceptions import InsecureRequestWarning
 
 # Disable SSL warnings when using verify=False
 urllib3.disable_warnings(InsecureRequestWarning)
+
+
+class DNSResolverAdapter(HTTPAdapter):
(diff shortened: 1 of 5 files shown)


tests/e2e/helper.py
Thanhphan1147, line 1: I don't think you need this, it complexifies the code a bit and you can always directly call https://<gateway_ip> and pass the correct host header. I would like us to start removing references to this dnsadapter

tests/e2e/helper.py
```diff
@@ -32,12 +100,19 @@ def assert_gateway_route_response(
     allow_redirects: bool = True,
 ) -> requests.Response:
     """Get a gateway route and assert expected response, retrying while dataplane converges."""
-    headers = {"Host": hostname} if hostname is not None else None
```
Thanhphan1147, line 35: I'm surprised that this doesn't work. Maybe I'm missing soemthing

f-atwi: but this is not about the host header which was not working
since the headers are encrypted and cilium would require the proper certificate to even read the header during TLS termination which it gets by matching using the SNI

Thanhphan1147 (approved):

PR #273, opened 2026-07-01
chore: test with juju4 (ISD-6047)

f-atwi:
#### Checklist

- [ ] I followed the [contributing guide](https://github.com/canonical/is-charms-contributing-guide)
- [ ] I updated `docs/changelog.md` with user-relevant changes
- [ ] I used AI to assist with preparing this PR
- [ ] I added or updated tests as needed (unit and integration)
- [ ] **If integration test modules are used:** I updated the workflow configuration  


spread.yaml
```diff
@@ -0,0 +1,72 @@
...
+    summary: cross-charm e2e tests (Juju 3)
+    working-dir: ./
+    discover-path: tests/e2e/
+    backends:
+    - integration-test
+    prepare: |
+      mkdir -p /home/ubuntu/.kube
+      k8s config > /home/ubuntu/.kube/config
+      chown ubuntu:ubuntu /home/ubuntu/.kube/config
+    pytest-arguments-template: '--kube-config=/home/ubuntu/.kube/config'
```
seb4stien: Isn't it the default value? Maybe we can just remove it.

gateway-api-integrator/tests/integration/test_charm_ingress_interface.py
```diff
@@ -42,7 +42,10 @@ def test_ingress_enforced_mode(
     gateway = get_gateway_resource(lightkube_client, application)
     gateway_lb_ip = gateway.status["addresses"][0]["value"]  # type: ignore
 
-    ingress_url = get_ingress_url_for_application(ingress_requirer_application, juju)
+    if juju.version().major >= 4:
+        ingress_url = get_ingress_url_for_application(configured_application_with_tls, juju)
+    else:
+        ingress_url = get_ingress_url_for_application(ingress_requirer_application, juju)
```
alithethird: would it be cleaner if we bake this switch logic into the `get_ingress_url_for_application` function? We are using the same logic whenever we call the function so it feels like this should be the responsibility of the function.

gateway-api-integrator/tests/integration/test_charm_dns.py
```diff
@@ -38,8 +39,14 @@ def test_dns_record_relation(
     juju.wait(jubilant.all_active)
 
     # Assert that the dns-record is in the relation data
-    unit_info_str = juju.cli("show-unit", "any-charm/0", "--format", "json")
-    unit_info_dict = json.loads(unit_info_str)["any-charm/0"]
+    if juju.version().major >= 4:
```
javierdelapuente, line 42: why don't we ask any-charm instead of this, or really test whatever what we are doing? 

I mean, it is not part of this PR, but I would change the approach instead.

gateway-api-integrator/tests/integration/test_charm_ingress_interface.py
```diff
@@ -42,7 +42,10 @@ def test_ingress_enforced_mode(
     gateway = get_gateway_resource(lightkube_client, application)
     gateway_lb_ip = gateway.status["addresses"][0]["value"]  # type: ignore
 
-    ingress_url = get_ingress_url_for_application(ingress_requirer_application, juju)
+    if juju.version().major >= 4:
+        ingress_url = get_ingress_url_for_application(configured_application_with_tls, juju)
```
javierdelapuente: What is the reason for this? why should it work differently in juju 4 and juju 3?

f-atwi: I wish I knew but this is what I wanted to discuss with the juju team
juju3 is providing the application data written to the application databag by the other app
in juju4, it is returning the application data the charm itself wrote

gateway-api-integrator/tests/integration/helper.py
```diff
@@ -119,18 +119,22 @@ def get_http_route_resource(
 
 
 def get_ingress_url_for_application(
-    ingress_requirer_application: str, juju: jubilant.Juju
+    ingress_requirer_application: str, ingress_provider_application: str, juju: jubilant.Juju
 ) -> ParseResult:
     """Get the ingress url from the requirer's unit data.
```
seb4stien, line 124: I would leave a comment on why we need this (as it may change in the future if that's not expected). Ideally, we should refer to a github issue.

seb4stien (approved):

PR #276, opened 2026-07-06
fix: reduce test flakiness

f-atwi:
#### What this PR does
Added delays to give the model enough time to converge.
Removes all but the e2e integration tests on juju3

Also changed the delay in juju.wait from 1s (default) to 5s to ensure we are not bypassing the wait condition with the stale active model from previous tests.

#### Why we need it

Juju4 is slow. 

#### Checklist

- [ ] I followed the [contributing guide](https://github.com/canonical/is-charms-contributing-guide)
- [ ] I updated `docs/changelog.md` with user-relevant changes
- [ ] I used AI to assist with preparing this PR
- [ ] I added or updated tests as needed (unit and integration)
- [ ] **If integration test modules are used:** I updated the workflow configuration  


gateway-api-integrator/tests/integration/conftest.py
```diff
@@ -18,7 +18,7 @@
 CERTIFICATE_PROVIDER_CHANNEL = "1/edge"
 INGRESS_REQUIRER_CHANNEL = "latest/edge"
 TEST_EXTERNAL_HOSTNAME_CONFIG = "gateway.internal"
-GATEWAY_CLASS_CONFIG = "cilium"
+GATEWAY_CLASS_CONFIG = "ck-gateway"
```
seb4stien, line 21: Is it supposed to be part of the PR?

f-atwi: ck-gateway is the default controller provided with ck8s. I thought might as well test with that instead of cilium.
Does not contribute to fixing flakiness yeah but probably better to keep it this way

seb4stien (approved):

PR #327, opened 2026-08-20
docs: add how-to guide for selecting a gateway class

f-atwi:
#### Checklist

### DIFF
--- a/docs/how-to/select-gateway-class.md
+++ b/docs/how-to/select-gateway-class.md
@@ -40,8 +40,16 @@ it goes into a blocked state whose status message lists the classes available:
 ```{terminal}
 :scroll:
 juju status
-App                     Version  Status   Scale  Charm                   Channel   Rev  Address         Exposed  Message
-gateway-api-integrator           blocked      1  gateway-api-integrator  1/stable  xxx  10.152.183.178  no       Gateway class must be one of: [cilium,ck-gateway]
+
+Model  Controller     Cloud/Region  Version  SLA          Timestamp
+test   concierge-k8s  k8s           3.6.27   unsupported  ...
+
+App                     Version  Status   Scale  Charm                   Channel  Rev  Address         Exposed  Message
+gateway-api-integrator           blocked      1  gateway-api-integrator  1/static xxx  10.152.183.178  no       Gateway class must be one of: [cilium,ck-gateway]
+
+Unit                       Workload  Agent  Address    Ports  Message
+gateway-api-integrator/0*  blocked   idle   10.1.0.44         Gateway class must be one of: [cilium,ck-gateway]
+
 ```
 
 ## Set the gateway class
--- a/docs/how-to/select-gateway-class.md
+++ b/docs/how-to/select-gateway-class.md
@@ -15,7 +15,7 @@ A `GatewayClass` is associated with a gateway controller (for example "Cilium")
 running on the Kubernetes cluster, so the value you choose must match a class that already exists on
 the cluster.
 
-```{note}
+```{caution}
 The `gateway-api-integrator` charm is currently tested against the `cilium` and `ck-gateway`
 gateway classes. Other gateway classes may work but are not verified.
 ```
--- a/docs/how-to/select-gateway-class.md
+++ b/docs/how-to/select-gateway-class.md
@@ -64,6 +64,7 @@ The blocked status message shows the valid options, for example:
 
 ```{terminal}
 :output-only:
+:scroll:
 App                     Version  Status   Scale  Charm                   Channel   Rev  Address         Exposed  Message
 gateway-api-integrator           blocked      1  gateway-api-integrator  1/stable  165  10.152.183.178  no       Gateway class must be one of: [cilium,ck-gateway]
 ```
--- a/docs/how-to/select-gateway-class.md
+++ b/docs/how-to/select-gateway-class.md
@@ -1,3 +1,9 @@
+---
+myst:
+  html_meta:
+    "description lang=en": "Learn how to set the gateway-class configuration for the gateway-api-integrator charm"
+---
+
 (how_to_select_gateway_class)=
 
 # How to select a gateway class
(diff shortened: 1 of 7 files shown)


docs/how-to/select-gateway-class.md
```diff
@@ -0,0 +1,66 @@
...
+```
+
+## List the available gateway classes
+
+The gateway classes available on a cluster are provided by the installed gateway controllers. List
+them with `kubectl`:
+
+```bash
+kubectl get gatewayclasses
+```
```
f-atwi: Should we discuss creating a `get-gateway-classes` juju action so that we can have a juju way of getting this info?

docs/how-to/select-gateway-class.md
```diff
@@ -0,0 +1,66 @@
...
+juju status gateway-api-integrator
+```
+
+The blocked status message shows the valid options, for example:
+
+```{terminal}
+:output-only:
+App                     Version  Status   Scale  Charm                   Channel   Rev  Address         Exposed  Message
+gateway-api-integrator           blocked      1  gateway-api-integrator  1/stable  165  10.152.183.178  no       Gateway class must be one of: [cilium,ck-gateway]
+```
```
f-atwi: does not look as intended on the rendered page

docs/how-to/select-gateway-class.md
```diff
@@ -0,0 +1,66 @@
+(how_to_select_gateway_class)=
```
erinecon, line 7: Let's also add an HTML metadata description:
```suggestion
---
myst:
  html_meta:
    "description lang=en": "Learn how to set the gateway-class configuration for the gateway-api-integrator charm"
---

(how_to_select_gateway_class)=
```
Feel free to edit as you see fit :)

docs/how-to/select-gateway-class.md
erinecon: I think you can use the `:scroll:` option so that the terminal won't wrap the long lines, see the [`sphinx-terminal` README](https://github.com/canonical/sphinx-terminal/blob/main/README.md) for more info

docs/how-to/select-gateway-class.md
erinecon: I think a new action makes sense *in theory*, but there are considerations that I'm not the best person to answer:
* How much time and effort would be involved in adding a new action?
* Are there any advantages to listing the gateway classes through a Juju action versus `kubectl`? (For instance: is `kubectl` a required component of the setup? Is `kubectl` well-known to charm operators?)
* What do you consider the end-user value to be for a dedicated `get-gateway-classes` action? (For instance, I could believe that a dedicated action would match users' expectations or mental models of charms more than using `kubectl`.)

docs/how-to/select-gateway-class.md
```diff
@@ -0,0 +1,66 @@
...
+[GatewayClass](https://gateway-api.sigs.k8s.io/reference/api-types/gatewayclass/)
+to use when creating the `Gateway` resource.
+A `GatewayClass` is associated with a gateway controller (for example "Cilium")
+running on the Kubernetes cluster, so the value you choose must match a class that already exists on
+the cluster.
+
+```{note}
+The `gateway-api-integrator` charm is currently tested against the `cilium` and `ck-gateway`
+gateway classes. Other gateway classes may work but are not verified.
+```
```
erinecon: Nit: Maybe we want to make this a `warning` or `caution` admonition instead

docs/how-to/select-gateway-class.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Set the gateway class
+
+Configure the charm with one of the available class names:
+
+```bash
+juju config gateway-api-integrator gateway-class=<gateway-class-name>
+```
+
+For example, to use the `ck-gateway` class shipped with Canonical Kubernetes:
```
erinecon:
```suggestion
For example, to use the `ck-gateway` class shipped with Canonical Kubernetes, run:
```
or

```suggestion
For example, use the `ck-gateway` class shipped with Canonical Kubernetes:
```

docs/how-to/select-gateway-class.md
erinecon, line 1: I think we should include this guide in the home page table! To determine whether we should use "Deployment" or "Operations" for the row title, let me ask you: Do you think this guide falls more under Day 0/1 operations, or Day 2+ operations? (If Day 0/1, I think the row should be named "Deployment". If Day 2+, then I think the row should be named "Operations").

erinecon (commented):
Thanks for adding this guide! Some initial thoughts from me

docs/how-to/select-gateway-class.md
f-atwi: Discussed this with the team and decided to reorder the checks to block the charm.
Now, the first guard would be to inform the user that the gateway-class is missing (taking precedence over missing certificates relation)
https://github.com/canonical/gateway-api-integrator-operator/pull/329

docs/how-to/select-gateway-class.md
f-atwi: Then I would add it to deployment as it is a necessary thing to configure during deployment

docs/index.md
```diff
@@ -20,6 +20,8 @@ This operator is built for the **Kubernetes** substrate.
   -
 * - **Get started**
   - {ref}`tutorial_getting_started` | {ref}`tutorial_using_gateway_route`
+* - **Deployment**
+  - {ref}`how_to_select_gateway_class`
```
erinecon:
```suggestion
  - {ref}`Select a gateway class <how_to_select_gateway_class>`
```
nit

docs/how-to/select-gateway-class.md
```diff
@@ -0,0 +1,70 @@
...
+---
+
+(how_to_select_gateway_class)=
+
+# How to select a gateway class
+
+The `gateway-class` configuration option tells the `gateway-api-integrator` charm which
+[GatewayClass](https://gateway-api.sigs.k8s.io/reference/api-types/gatewayclass/)
+to use when creating the `Gateway` resource.
+A `GatewayClass` is associated with a gateway controller (for example "Cilium")
```
erinecon:
```suggestion
A `GatewayClass` is associated with a gateway controller
```
nit, IMO it's not necessary to include this example in the intro

erinecon (approved):
Your decisions and changes LGTM, thanks so much for adding this guide 🎉 Some non-blocking nits remain on my side

